PR #68, opened 2020-11-24
Added new shortcut to open full images 'V'. Changed behavior to open new tab.

yoditax:
### Before you begin

---

### DIFF
--- a/assets/js/shortcuts.js
+++ b/assets/js/shortcuts.js
@@ -13,7 +13,7 @@ function openFullView() {
   const imageHover = $('[data-uris]:hover');
   if (!imageHover) return;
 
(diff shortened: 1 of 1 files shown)


WingbeatPony:
Open in new tab behavior breaks in Firefox if popup-blocking behavior is enabled, rendering this shortcut useless. The button this shortcut emulating does not behave this way, either, and no other link or function on the site opens a new tab.

Additionally, this shortcut only works on image pages if you are hovering on the image, as though it were a thumbnail - unlike other image page shortcuts such as favorite and upvote.

Having two separate shortcuts for the same functionality (O and V) is also unnecessary, and it might be better if one of these two shortcuts (V) were repurposed on the image page to instead trigger the function to resize the image.

PR #133, opened 2021-09-05
Added 'Create Gallery' button to Galleries page.

Erhannis:
Dunno if it's in the right spot or whatever, and the 'Create Gallery' page isn't really helpful if you don't already know what image to use as a cover, but at least this gets their foot in the door.

### Before you begin

---

I copied bits of code into a different place and trimmed them down until it made a passable "Create Gallery" button.  :P  Let me know if it should look different or be in a different place or anything.


Erhannis:
Oh - this is for https://github.com/derpibooru/philomena/issues/216

PR #182, opened 2023-05-03
feat: add ability for staff to delete tag changes from tag change history

Neetpone:
### Before you begin

---

This used to be a thing in Booru on Rails but wasn't added to Philomena for some reason.

### DIFF
--- a/lib/philomena_web/controllers/image/tag_change_controller.ex
+++ b/lib/philomena_web/controllers/image/tag_change_controller.ex
@@ -60,14 +60,14 @@ defmodule PhilomenaWeb.Image.TagChangeController do
 
   defp log_details(conn, _action, %{image: image, details: details}) do
     %{
-      body: "Deleted tag change (#{details}) from history",
+      body: "Deleted tag change #{details} on >>#{image.id} from history",
       subject_path: Routes.image_path(conn, :show, image)
     }
   end
 
   defp tag_change_details(%TagChange{added: true, tag: tag}),
-    do: "added tag \"#{tag.name}\""
+    do: "+#{tag.name}"
 
   defp tag_change_details(%TagChange{added: false, tag: tag}),
-    do: "removed tag \"#{tag.name}\""
+    do: "-#{tag.name}"
 end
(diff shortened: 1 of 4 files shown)


liamwhite (approved):

lib/philomena_web/controllers/image/tag_change_controller.ex
```diff
@@ -35,4 +57,17 @@ defmodule PhilomenaWeb.Image.TagChangeController do
...
+
+  defp log_details(conn, _action, %{image: image, details: details}) do
+    %{
+      body: "Deleted tag change (#{details}) from history",
+      subject_path: Routes.image_path(conn, :show, image)
+    }
+  end
+
+  defp tag_change_details(%TagChange{added: true, tag: tag}),
+    do: "added tag \"#{tag.name}\""
```
Meow, line 69: maybe instead of `added/removed tag "tagname"` simply put it like `+tagname` or `-tagname`?

lib/philomena_web/controllers/image/tag_change_controller.ex
```diff
@@ -35,4 +57,17 @@ defmodule PhilomenaWeb.Image.TagChangeController do
 
   defp added_filter(query, _params),
     do: query
+
+  defp log_details(conn, _action, %{image: image, details: details}) do
+    %{
+      body: "Deleted tag change (#{details}) from history",
```
Meow, line 63: maybe also include which image ID is affected in the string?
`Deleted tag change +/-tagname on image 12345 from history.`

lib/philomena_web/controllers/image/tag_change_controller.ex
AppleDash: I was wondering about this, I considered this but I wasn't sure if it was a good idea or not, glad to hear that it is indeed a good idea - done!

lib/philomena_web/controllers/image/tag_change_controller.ex
AppleDash: The reason I didn't do this is because the mod log entry always links to the image page, but I saw now that it's like this on other mod log entries, so I might as well be consistent! Did it!

PR #205, opened 2024-03-06
Make API key & RSS feed url only show when clicked

EXtremeExploit:
### Before you begin

---
Having the API key show in plain sight in the account page is bad practice, there would be risks of people accidentally leaking their API keys because they went to the account page to check something but showing the API key to the public in the process in the case they were sharing their screens or streaming  on twitch.
Currently the way i am doing this behaviour is super scuffed, having the box be treated as an image filter is super wrong but its the only way i could think of. Same goes for overriding the initial pass of actions possible to the element which were "unspoilering" the api key.
If theres a better way to do this please tell so, i am very new to elixir/slime/this software, thanks!

### DIFF
--- a/lib/philomena_web/templates/registration/edit.html.slime
+++ b/lib/philomena_web/templates/registration/edit.html.slime
@@ -20,9 +20,13 @@ p
 h3 API Key
 p
   ' Your API key is
-  code>
-    = @current_user.authentication_token
-  ' - you can use this to allow API consumers to access your account.
+  #api-key-button>
+    code>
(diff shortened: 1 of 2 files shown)


liamwhite:
I would suggest using the data-click-show="#target" attribute for this instead of hacking it to use image templates. Also, you should probably also do this with the RSS endpoint shown in the settings as it uses the same API key.

EXtremeExploit:
Thanks for the suggestion, this is how it looks now:
image
image

PR #242, opened 2024-04-30
Vite HMR for JS/TS (and jankily for CSS)

Meow:
HMR aka vite's screwed up definition of live reload. Runs two vite instances (BECAUSE YAY), one just dumbly compiles static bundle for fonts and fallback theme to prevent FOUC, and the other is the dev server that does LiVe ReLoAdInG ThInGy because it's 2024 and apparently live reloading is all the rage so we cheaply as possible cosplay one of those fancy fabled SINGLE PAGE APPLICATION FRAMEWORKS because that's cool these days I guess

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -8,3 +8,6 @@
 // Our code
 import './ujs';
 import './when-ready';
+
+// Fallback on the dark theme for now
+import '../css/themes/dark.scss';
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -9,5 +9,11 @@
 import './ujs';
 import './when-ready';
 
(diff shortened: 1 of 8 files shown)


mdashlw:
After this PR, it is now required to disable ublock origin on localhost when running in dev. uBlock blocks http://localhost:5173/js/fingerprint.js which is not handled and causes most js to fail as well.

Meow:
> After this PR, it is now required to disable ublock origin on localhost when running in dev. uBlock blocks http://localhost:5173/js/fingerprint.js which is not handled and causes most js to fail as well.

Sounds like a uBlock bug (why would it block stuff on localhost)

PR #339, opened 2024-07-22
use keyCode instead of key

mdashlw:
### Before you begin

---

`keyCode` may be deprecated but it provides a more convenient behavior regarding different keyboard layouts


Meow:
I disagree on using deprecated API, maybe it's possible to find a non-deprecated workaround

mdashlw:
> I disagree on using deprecated API, maybe it's possible to find a non-deprecated workaround

I think reimplementing `keyCode` is a waste of time and would be prone to error. It just works and is unlikely to ever be dropped.

PR #409, opened 2025-02-10
Display alias names for the aliases matched by local autocomplete

koloml:
### Before you begin

---

All aliases would be displayed in the following format:

```
alias ⇒ tag name (count)
```

This change is only updates how aliases displayed from the local autocompleter, server-side suggestions are not updated.

Aliased tags from Furbooru

### DIFF
--- a/assets/js/utils/suggestions.ts
+++ b/assets/js/utils/suggestions.ts
@@ -1,7 +1,7 @@
 import { makeEl } from './dom.ts';
 import { mouseMoveThenOver } from './events.ts';
 import { handleError } from './requests.ts';
-import { LocalAutocompleter } from './local-autocompleter.ts';
+import { LocalAutocompleter, Result } from './local-autocompleter.ts';
 
 export interface TermSuggestion {
   label: string;
@@ -174,4 +174,17 @@ export async function fetchLocalAutocomplete(): Promise<LocalAutocompleter> {
     .then(handleError)
     .then(resp => resp.arrayBuffer())
     .then(buf => new LocalAutocompleter(buf));
+}
+
+export function formatLocalAutocompleteResult(result: Result): TermSuggestion {
+  let tagName = result.name;
+
+  if (tagName !== result.aliasName) {
+    tagName = `${result.aliasName} ⇒ ${tagName}`;
+  }
+
+  return {
+    value: result.name,
+    label: `${tagName} (${result.imageCount})`,
+  };
 }
--- a/assets/js/autocomplete.ts
+++ b/assets/js/autocomplete.ts
@@ -7,7 +7,13 @@ import { getTermContexts } from './match_query';
 import store from './utils/store';
 import { TermContext } from './query/lex';
 import { $$ } from './utils/dom';
-import { fetchLocalAutocomplete, fetchSuggestions, SuggestionsPopup, TermSuggestion } from './utils/suggestions';
+import {
+  formatLocalAutocompleteResult,
+  fetchLocalAutocomplete,
+  fetchSuggestions,
+  SuggestionsPopup,
+  TermSuggestion,
+} from './utils/suggestions';
 
 let inputField: HTMLInputElement | null = null,
   originalTerm: string | undefined,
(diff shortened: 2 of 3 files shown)


assets/js/utils/suggestions.ts
```diff
@@ -175,3 +175,16 @@ export async function fetchLocalAutocomplete(): Promise<LocalAutocompleter> {
     .then(resp => resp.arrayBuffer())
     .then(buf => new LocalAutocompleter(buf));
 }
+
+export function formatLocalAutocompleteResult(result: Result): TermSuggestion {
+  let tagName = result.name;
+
+  if (tagName !== result.aliasName) {
+    tagName = `${result.aliasName} ⇒ ${tagName}`;
```
liamwhite, line 183: I believe it would be more clear if this was the other way around (so that the prefix of what you typed appears first, followed by what it is aliased to)

MareStare: I suppose `artist:assasinmoney` has an alias `test` in the example. But why would the autocompleter  event suggest `test` alias at all if the query matches the `artist:assasinmoney` directly?

I'd expect this:

- Input: `as`. Results: 
  - `artist:assasinmoney`
  
- Input: `te`. Results:
  - `test => artist:assasinmoney`. Pressing `Enter` autocomplete `artist:assasinmonkey`

liamwhite: I believe the alias direction is artist:assasinmonkey => test in the PR example (this is implied by the typed prefix), which is why I left my comment.

MareStare: I'd expect the `artist` tag to be the canonical tag name. So which one is the canonical name in the example?

koloml: The screenshot is kind of misleading, since I made up the alias on the spot with manual code change. So the actual behavior will be:

- When `test` is typed, it would display `test => artist:assasinmoney (1)`
- When `assasinmoney` is typed, it would display just the `assasinmoney (1)`.

koloml: But I'm personally indifferent to the order of the tag & alias — not really used to how other engines display aliases. It could be any of these three:

- `alias ⇒ tag (count)`
- `tag ⇒ alias (count)`
- `tag ⇐ alias (count)`

MareStare: I suggest to establish two properties here:
- The arrow should point in the direction of the canonical tag name suggesting to use the canonical representation instead.
- The tersm that matche the input query should be consistently visually aligned. I.e. it will be confusing if the matched terms occur in different parts of the pairs:
   ```
   Query: foo
   foo => bar
   something => foobaz
   foox
   fooabc => zoom
   bleh => fooing
   ```
   Instead, I think such a query just shouldn't show the aliases `something` and `bleh` (remember the arrow points in the direction of the canonical tag). It should instead be:
   ```
   Query: foo
   foo => bar
   foobaz
   foox
   fooabc => zoom
   fooing
   ```
   
I assume current tag matching logic doesn't return aliases of the canonical tag name. For example, if you type `twi`, the results will list just `twilight sparkle` and not all of her aliases (`twilightsparkle`, `twilight sparkles`, etc.).

koloml: > I assume current tag matching logic doesn't return aliases of the canonical tag name. For example, if you type twi, the results will list just twilight sparkle and not all of her aliases (twilightsparkle, twilight sparkles, etc.).

Yes, local autocompleter only provides aliases when we're actively matching them, not the other way around.

So the current implementation will show matched aliases on the left and the original tags on the right. So if we will search for `foo` and there would be alias on the `foo` to `bar` but there will be no aliases on the `foobaz` then suggestions would be:

```
foo => bar (12)
foobaz (10)
```

And if we will search for `bar`, then there would be only `bar`:

```
bar (12)
```

koloml: Actually, here is live example from Furbooru, where `whfr` is aliased to `white fur`:

image

But if you type `white fu` then aliases won't be displayed:

image

PR #412, opened 2025-02-11
Allow regular users to see duplicate images

mdashlw:
### Before you begin

---

Allows everyone to `:show` and `:index` duplicate images. All other interaction is already blocked by handlers below.
Except for `:vote`, which relies on `:show`, so currently users will be allowed to vote on duplicate images.

The default behavior does not change. Normal users will still be redirected to target image *unless* you add `?del=1` to image url. Mods/admins will not be redirected regardless.

Changes `:hide` to `:show` in a couple places, not sure what the implications of that are.

API will return duplicate images by default with `hidden_from_users: true` and `duplicate_of` which should not make this a breaking change.
A side-effect of API now doing `can?(user, :show, image)` is that mods/admins can now use their api key and view (non-duplicate) deleted images.

### DIFF
--- a/lib/philomena/users/ability.ex
+++ b/lib/philomena/users/ability.ex
@@ -386,6 +386,10 @@ defimpl Canada.Can, for: [Atom, Philomena.Users.User] do
       when action in [:show, :index],
       do: true
 
+  def can?(_user, action, %Image{hidden_from_users: true, duplicate_id: duplicate_id})
+      when action in [:show, :index] and not is_nil(duplicate_id),
+      do: true
+
   def can?(_user, :show, %Tag{}), do: true
 
   # Comment on images where that is allowed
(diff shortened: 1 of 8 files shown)


Meow:
I am not entirely sure if that's something that is desired, or if there are any edge cases or security implications to this (merging a duplicate DOES delete the image and assign a hidden image key to it, so its version looks like 12345-deadbeef123 or something, so that the old image is not viewable by normal means). Usually it'd just redirect a normal user to a merged image.

Mayhaps if this is implemented, we could avoid moving images to hidden image key version altogether and just keep it in place? And for existing images which already have been moved to a hidden image key, put in place some sort of code that shows that version instead? e.g. `if img.hidden_from_users && img.duplicate_of != nil && img.hidden_image_key != nil` (pseudocode)

Meow:
Not something we want. However, I'd accept a different PR for a feature to let trusted users (via a special user role) see that information, exactly like implemented here. Just gate it behind a special role (similar to how batch tagging is gated), without impacting regular user's experience. Also make users with that special role see the image like mods, aka without redirect.

PR #425, opened 2025-02-21
move communication text below avatar on mobile

Meow:
mobile communications less sucky wucky uwu

### DIFF
--- a/assets/css/views/communications.css
+++ b/assets/css/views/communications.css
@@ -158,12 +158,18 @@ span.communication__sender__stats,
 
   .communication__body__avatar {
     grid-area: 1 / 1 / 2 / 2;
+    align-self: center;
   }
 
   .communication__body__sender {
     /* 8px body padding + 2px borders + 12px article padding + 12px avatar margin */
     width: calc(100vw - 34px - var(--reduced-communication-avatar-size));
     grid-area: 1 / 2 / 2 / 3;
+    padding-bottom: 6px;
+  }
+
+  .communication__body__sender > .label--block {
+    margin-bottom: 0;
   }
 
   .communication__body__text {
--- a/assets/css/views/communications.css
+++ b/assets/css/views/communications.css
@@ -113,6 +113,37 @@ span.communication__sender__stats,
   margin-right: 12px;
 }
 
+/* Override padding from block__content, so we can do it on a per-element basis */
+.communication__content {
+  padding: 0;
+}
+
+.communication__body__avatar,
+.communication__body__sender,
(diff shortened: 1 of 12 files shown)


assets/css/views/communications.css
```diff
@@ -107,3 +107,27 @@ span.communication__sender__stats,
...
+/* Mobile layout fixes */
+@media (max-width: $min-desktop-width) {
+  .communication__body {
+    display: grid;
+    grid-template-columns: repeat(2, auto) repeat(3, 1fr);
+    grid-template-rows: repeat(2, auto) repeat(3, 1fr);
+  }
+
+  .communication__body:first-child {
+    grid-area: 1 / 1 / 2 / 2;
```
liamwhite, line 120: Use an actual class name for this

liamwhite (dismissed):
For reference, it looks like this

image

There's no visual distinction (like a background color change) between the avatar row and the communication body row which is unpleasant

Also, the avatar is huge, and wasting tons of vertical space on a presumably high dpi screen

Meow:
> For reference, it looks like this
> 
> image
> 
> There's no visual distinction (like a background color change) between the avatar row and the communication body row which is unpleasant
> 
> Also, the avatar is huge, and wasting tons of vertical space on a presumably high dpi screen

The goal of this PR is to simply move the next below avatar + name block, nothing else. tl;dr i am not doing this in this PR, do not ask

liamwhite (approved):

assets/css/views/communications.css
```diff
@@ -107,3 +107,86 @@ span.communication__sender__stats,
...
+  flex: 0 0 auto;
+  margin-right: 12px;
+}
+
+/* Override padding from block__content, so we can do it on a per-element basis */
+.communication__content {
+  padding: 0;
+}
+
+.communication__body__avatar,
```
Veetaha, line 121: I think sub-elements are prohibited by BEM, and I was not happy to figure that out. But I see that this rule isn't followed intentionally in this codebase?

MareStare, line 121: I think sub-elements (i.e. several __) are [prohibited](https://en.bem.info/methodology/quick-start/#nesting-1) by BEM:
> An element is always part of a block, not another element. This means that element names can't define a hierarchy such as block__elem1__elem2.

I was not happy to figure that out. But I see that this rule isn't followed intentionally in this codebase?

liamwhite: Indeed the codebase was adapted into it and it was never really followed very closely (tons of nested rules...)

PR #427, opened 2025-02-22
Fix Canada checks for users with multiple roles of the same resource type

mdashlw:
### Before you begin

---

### DIFF
--- a/lib/philomena/users.ex
+++ b/lib/philomena/users.ex
@@ -943,7 +943,10 @@ defmodule Philomena.Users do
   defp setup_roles(nil), do: nil
 
   defp setup_roles(user) do
-    role_map = Map.new(user.roles, &{&1.resource_type || &1.name, &1.name})
+    role_map =
+      user.roles
+      |> Enum.group_by(& &1.resource_type, & &1.name)
+      |> Map.new(fn {type, names} -> {type, Map.new(names, &{&1, []})} end)
 
     %{user | role_map: role_map}
   end
--- a/lib/philomena_web/templates/admin/user/_form.html.slime
+++ b/lib/philomena_web/templates/admin/user/_form.html.slime
@@ -39,10 +39,6 @@
   .block
     .block__header
       span.block__header__title General user flags
-    ul
-      p
-        strong> Be careful when issuing these permissions to staff members!
-        | Staff members with relevant roles may have these permissions anyway. Issuing them may break things, such as tag aliasing.
     ul = collection_checkboxes f, :roles, filtered_roles(general_permissions(), @roles), mapper: &checkbox_mapper/6
 
   .block
--- a/lib/philomena/users/ability.ex
+++ b/lib/philomena/users/ability.ex
@@ -46,7 +46,7 @@ defimpl Canada.Can, for: [Atom, Philomena.Users.User] do
   def can?(%User{role: "moderator"}, :show, %Filter{}), do: true
 
   # Privileged mods can hard-delete images
-  def can?(%User{role: "moderator", role_map: %{"Image" => "admin"}}, :destroy, %Image{}),
+  def can?(%User{role: "moderator", role_map: %{"Image" => %{"admin" => _}}}, :destroy, %Image{}),
     do: true
 
   # ...but normal ones cannot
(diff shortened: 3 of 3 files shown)


liamwhite (approved):

lib/philomena/users.ex
```diff
@@ -943,7 +943,10 @@ defmodule Philomena.Users do
   defp setup_roles(nil), do: nil
 
   defp setup_roles(user) do
-    role_map = Map.new(user.roles, &{&1.resource_type || &1.name, &1.name})
+    role_map =
+      user.roles
+      |> Enum.group_by(& &1.resource_type, & &1.name)
```
Veetaha, line 948: Previously the expression was `&1.resource_type || &1.name` meaning it was handling the case where `&1.resource_type` could be `nil`. I suppose it never is `nil`? Becase defaulting `resource_type` to `name` was strange. Just want to confirm that non-`nil` assumption in new code is correct.

lib/philomena_web/templates/admin/user/_form.html.slime
```diff
@@ -39,10 +39,6 @@
   .block
     .block__header
       span.block__header__title General user flags
-    ul
-      p
-        strong> Be careful when issuing these permissions to staff members!
```
Veetaha, line 44: What was the problem here before with breaking tag aliasing? Could you recall the details?

lib/philomena/users.ex
```diff
@@ -943,7 +943,10 @@ defmodule Philomena.Users do
...
+      |> Map.new(fn {type, names} -> {type, Map.new(names, &{&1, []})} end)
```
Veetaha, line 949: This empty list dummy value in the map looks a bit weird, could use a `MapsSet` instead, but I get that this way it's easier to pattern-match in canada checks.

Maybe instead of using a `Map` here we could just keep a scalar role name here by finding the maximum from the role names. For example `["moderator", "admin"]` should coerce to just `"admin"` and `["moderator"]` should coerce to just `"moderator"`. Could use an `Enum.max_by` for that. This way we won't need any change in `ability.ex`.

lib/philomena/users.ex
mdashlw: `resource_type` shouldn't be nil although it's missing a db constraint. I mentioned this in a cleanup pr.

lib/philomena_web/templates/admin/user/_form.html.slime
mdashlw: If you had `Tag:batch_update` it'd override `Tag:moderator` and `Tag:admin` which give more permissions

lib/philomena/users.ex
mdashlw: I thought of this and I don't think this is a fair assumption for this system

lib/philomena_web/templates/admin/user/_form.html.slime
MareStare: Interesting. So this was the manifestation of the bug fixed by this PR, right? But I don't understand why a note was added here on UI? That note was basically saying "we have a bug with Tag:batch_update overriding Tag:moderator/admin, so please don't combine them". Rather than adding that note it would've been better to fix the bug. 

Or maybe I'm misinterpreting that note?

MareStare (approved):
LGTM, @mdashlw let's merge this unless you want to do any last-minute changes

lib/philomena_web/templates/admin/user/_form.html.slime
mdashlw: The note was about the bug fixed by this PR, yes.

lib/philomena/users.ex
mdashlw: The permission model is a little scuffed because some places bypass Canada and depend on user's role. And "roles" are more capabilities than roles, yes. But if it works it works. We don't really need complicated access control.

MareStare:
# QA
Validated some simple scenarios, found no regressions. From the quirks - it's possible to assign moderator/admin level special privileges even for regular users, although they won't take effect. That is rather confusing, but I think we'll tackle that as part of v2

PR #448, opened 2025-03-04
[Part 11] Add `AutocomplableInput` model for UI parsing

MareStare:
### Before you begin

---

This component is responsible for parsing the autocompletable input's data attributes and lexing the user query.

### Data Attribute Config

Search history can be enabled for certain `<input>` or `<textarea>` elements with the new data attribute `data-autocomplete-history-id={history_id}`. The `{history_id}` should be some human-readable word that will be used as a prefix in the local storage key, which is formatted as `{history_id}-history`.

Currently, only the main search input in the header page and this auxiliary input on the images `/search` page use this feature. Several inputs can share the same search history if they use the same `history_id`. I assigned `history_id = 'search'` to both, and it results in a `search-history` key in `localStorage`.

### DIFF
--- a/assets/js/autocomplete/input.ts
+++ b/assets/js/autocomplete/input.ts
@@ -1,0 +1,199 @@
+import store from '../utils/store';
+import { getTermContexts } from '../match_query';
+import { Range } from '../query/lex';
+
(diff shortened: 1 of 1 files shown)


liamwhite (approved):

MareStare:

PR #451, opened 2025-03-04
[Part 14] Redesign the autocomplete suggestions popup

MareStare:
### Before you begin

---

## UI Changes

Updated the UI design of the completions substantially. Here are some notable differences.

### Alias Arrow

I replaced the fat arrow character `⇒` with a thin arrow `→` for aliases, and to make that work I changed the font to the monospace font used in other places on the frontend.

image

### Image Count Rendering

I've decided to format the number in full with groups of 3 digits separated by a short space. I think this way of rendering looks nicer than using `M / K` contractions, since this way all the digits are aligned and it's easy to see the difference in the order of numbers between different tags. I think this way of formatting scales well for the observable future of Derpibooru. Also, this way of displaying the image count will make Derpibooru's completions UI style more unique (less of a copy-paste from Danbooru).

### Mouseover

Hovering over the item in the completions list doesn't move the selection. Instead, it renders a visual-only highlight. This behaviour was borrowed from VSCode's completion behaviour. The left side is the old implementation and the right side is the new implementation:

https://github.com/user-attachments/assets/88f0b047-7cb2-400c-81c1-228616bbedea

### Navigation

- <kbd>Ctrl+Down</kbd>  - Jump to the next block of suggestions. This is intended to skip history suggestions and jump to the first tag suggestion.

- <kbd>Ctrl+Up</kbd> - Jump to the prev block of suggestions. E.g. if you are in the tag suggestions, that will jump to the last history suggestion. I don't see a real use case for this key binding, added it just for symmetry.

## Unresolved Questions

- Is the [shortcuts' documentation](https://derpibooru.org/pages/shortcuts) page source somewhere in public? If not you should edit it to add some notes about the autocompletion shortcuts mentioned above.

## UI Extremes

Overly long suggestions are truncated with an ellipsis. The overall max width of the completions box can be at most 70% of the viewport. This solution doesn't work all of the time (e.g. when the suggestions are very close to the right side of the screen) but it will work in most of the cases.

Here is the demo of handling extreme suggestion/image count lengths:

https://github.com/user-attachments/assets/23056bbc-fedb-489a-ba06-45214698a028

Note that the position of the completions box stays static when the browser is resized. It also works this way in the implementation in master, and I don't think it's a big issue. The box is rerendered the next time the user types something. I haven't found a simple solution for how to fix this yet.

### DIFF
--- a/assets/js/utils/suggestions.ts
+++ b/assets/js/utils/suggestions.ts
@@ -64,13 +64,11 @@ export class TagSuggestion {
   }
 
   static formatImageCount(count: number): string {
-    const chars = [...count.toString()];
+    // We use the 'fr' (French) number formatting style with space-separated
+    // groups of 3 digits.
+    const formatter = new Intl.NumberFormat('fr', { useGrouping: true });
 
-    for (let i = chars.length - 3; i > 0; i -= 3) {
-      chars.splice(i, 0, ' ');
-    }
-
-    return chars.join('');
+    return formatter.format(count);
   }
 }
 
--- a/assets/js/utils/suggestions.ts
+++ b/assets/js/utils/suggestions.ts
@@ -149,7 +149,6 @@ export class SuggestionsPopup {
       tabIndex: -1,
     });
 
-    // Make the container connected to DOM to make sure it's rendered when we unhide it
     document.body.appendChild(this.container);
     this.items = [];
   }
--- a/assets/js/utils/suggestions.ts
+++ b/assets/js/utils/suggestions.ts
@@ -1,125 +1,335 @@
 import { makeEl } from './dom.ts';
-import { mouseMoveThenOver } from './events.ts';
-import { handleError } from './requests.ts';
-import { LocalAutocompleter, Result } from './local-autocompleter.ts';
 
-export interface TermSuggestion {
-  label: string;
-  value: string;
+export interface TagSuggestionParams {
+  /**
+   * If present, then this suggestion is for a tag alias.
+   * If absent, then this suggestion is for the `canonical` tag name.
+   */
+  alias?: null | string;
+
+  /**
+   * The canonical name of the tag (non-alias).
+   */
+  canonical: string;
+
+  /**
+   * Number of images tagged with this tag.
+   */
+  images: number;
+
+  /**
+   * Length of the prefix in the suggestion that matches the prefix of the current input.
+   */
+  matchLength: number;
 }
 
-const selectedSuggestionClassName = 'autocomplete__item--selected';
+export class TagSuggestion {
+  alias?: null | string;
+  canonical: string;
+  images: number;
+  matchLength: number;
 
+  constructor(params: TagSuggestionParams) {
+    this.alias = params.alias;
+    this.canonical = params.canonical;
+    this.images = params.images;
+    this.matchLength = params.matchLength;
+  }
+
+  value(): string {
+    return this.canonical;
+  }
+
+  render(): HTMLElement[] {
+    const { alias: aliasName, canonical: canonicalName, images: imageCount } = this;
+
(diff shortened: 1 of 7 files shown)


assets/js/utils/suggestions.ts
```diff
@@ -1,195 +1,351 @@
...
+
+  static formatImageCount(count: number): string {
+    const chars = [...count.toString()];
+
+    for (let i = chars.length - 3; i > 0; i -= 3) {
+      chars.splice(i, 0, ' ');
+    }
+
+    return chars.join('');
+  }
```
liamwhite, line 74: maybe consider using [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) instead?

assets/js/utils/suggestions.ts
```diff
@@ -1,195 +1,351 @@
...
     this.container = makeEl('div', {
-      className: 'autocomplete',
+      className: 'autocomplete hidden',
+      tabIndex: -1,
     });
 
-    this.listElement = makeEl('ul', {
-      className: 'autocomplete__list',
-    });
+    // Make the container connected to DOM to make sure it's rendered when we unhide it
```
liamwhite, line 152: This comment is confusing me. Doesn't it just need to be in the DOM, period, because it's expected to be user-visible?

assets/js/utils/suggestions.ts
```diff
@@ -1,195 +1,351 @@
...
 
-  private watchItem(listItem: HTMLElement, suggestion: TermSuggestion) {
-    // This makes sure the item isn't selected if the mouse pointer happens to
-    // be right on top of the item when the list is rendered. So, the item may
-    // only be selected on the first `mousemove` event occurring on the element.
-    // See more details about this problem in the PR description:
-    // https://github.com/philomena-dev/philomena/pull/350
-    mouseMoveThenOver(listItem, () => this.updateSelection(listItem));
+  appendSuggestion(suggestion: Suggestion) {
+    const type = suggestion instanceof TagSuggestion ? 'tag' : 'history';
```
liamwhite, line 225: Is it possible to avoid the use of instanceof here, perhaps the caller is aware of the type?

MareStare: The `Suggestion` type is basically a discriminated union and it's used as such in here as well in PR https://github.com/philomena-dev/philomena/pull/453:
image.

I considered adding a `type` field or method to both the `TagSuggestion` and `HistorySuggestion` class objects, but then I thought it would be redundant since we already have the class prototype as a discriminator. I don't want the caller to pass an additional parameter to specify the type of the suggestion and thus decoupling the discriminator from the union which loses some type safety.

Would you prefer to keep it with an instanceof check or with adding `type = 'history'/'tag'` fields/methods to `TagSuggestion/HistorySuggestion` classes?

assets/js/utils/suggestions.ts
MareStare: Yeah, maybe this comment is just too obvious. I wrote it because prev. implementation didn't instantly connect the element to the dome. Given how different the new impl, I'll remove this one

assets/js/utils/suggestions.ts
MareStare: Hm, I'll take a look. Would be awesome if it implements the three-digit spacing for us

MareStare: I'll take a look. Would be awesome if it implements the three-digit spacing for us

MareStare: Tried french formatting and it yielded a "[narrow no-break space](https://symbl.cc/en/202F/)". I'll try this one and remove the manual `-3px` in CSS then
image

MareStare:

PR #464, opened 2025-03-18
Hotfix the color styling and fonts for autocompletions

MareStare:
### Before you begin

---

Changed the selection colours on dark and light themes to use their colours instead of black/white. The stronger highlight (higher) is used for the keyboard selection and a smaller highlight (lower) is for mouse on-hover on the screenshots below.

What isn't done in this PR is that I haven't switched the layout to a table yet. Will do it tomorrow.

### Before (Dark)
image

### After (Dark)
image

### Before (Light)
The mouse on-hover was absolutely invisible:

image

### After (Light)

image

## Font Handling

Added `Consolas` and some `Noto Sans Mono` that was used in Liam's browser in hope of getting a better monospace font by default. But even if the browser falls back to `Courier New`, the font will still look fine, because I made it fully `bold`. This fixes the following issue with `Courier New`, where the font is too thin, and becomes hardly readable on some sreens.

This is how it looks on my screen before the change:
image

This is how it looked in [this forum post](https://derpibooru.org/forums/meta/topics/site-development-notification-and-feedback-thread?post_id=5706601#post_5706601): 

image

Now it'll look like this:
image

## Pre-computing Relative Color 

As some anon mentioned on [the forum](https://derpibooru.org/forums/meta/topics/site-development-notification-and-feedback-thread?post_id=5706671#post_5706671), the relative CSS colour is too new (in fact it's in [working draft state](https://caniuse.com/?search=relative%20color)), so it wasn't working in some browsers. The workaround is to pre-compute the relative colour with postCSS.

For some reason, I didn't question whether the relative color in CSS was widely available or not, it was just convenient to use it and tweak some constants in the expression live in the browser, but unfortunately, that's too big of a luxury =).

## Better colour tuning

I manually added some color tuning for every available theme for the history suggestion items to improve the contrast for each of them.

---

Also reduced the spacing between the digits in the tag image count manually via `word-spacing`. Turns out that "narrow no-break space" doesn't actually render narrower for me.

### DIFF
--- a/assets/css/themes/base/dark.css
+++ b/assets/css/themes/base/dark.css
@@ -66,7 +66,9 @@
     --block-header-link-text-hover-color: hsl(from $block-header-link-text-color calc(h + 6) calc(s - 20) calc(l - 3));
 
     --block-header-light-hover-color: hsl(from $block-header-light-color calc(h - 4) calc(s + 10) calc(l - 4));
-    --block-header-light-link-text-hover-color: hsl(from $block-header-light-link-text-color calc(h + 8) calc(s - 10) calc(l - 2));
+    --block-header-light-link-text-hover-color: hsl(
+      from $block-header-light-link-text-color calc(h + 8) calc(s - 10) calc(l - 2)
+    );
 
     --media-box-hover-color: hsl(from $media-box-color h s calc(l - 4));
     --media-box-header-link-text-hover-color: hsl(from $link-color h calc(s - 18) calc(l - 3));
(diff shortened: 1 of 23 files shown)


assets/js/utils/suggestions.ts
```diff
@@ -68,7 +68,9 @@ export class TagSuggestion {
     // groups of 3 digits.
     const formatter = new Intl.NumberFormat('fr', { useGrouping: true });
 
-    return formatter.format(count);
+    // Normalize the whispace with a `.replace()`. We'll use CSS to guarantee
```
liamwhite, line 71:
```suggestion
    // Normalize whitespace with `.replace()`. We'll use CSS to guarantee
```

MareStare:
Forgot to attach an example of the new number rendering:

Consolas:
image

Courier New:
image

liamwhite (approved):

PR #472, opened 2025-03-23
Add init script and instructions for dev setup

MareStare:
### Before you begin

---

Added a helpful script for initial dev setup and some instructions to the README.md

Plus fixed the pre-commit hook to auto-format the Elixir code instead of failing with an error if Elixir code is not formatted.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# VSCode supports `jsonc` for all of its JSON files
+.vscode/*.json linguist-language=jsonc
+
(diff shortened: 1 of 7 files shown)


mdashlw:
vscode extensions could be added this way: https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

Veetaha:
Oh yeah, totally forgot about it, will add a list

MareStare:
Oh yeah, totally forgot about it, will add a list

PR #473, opened 2025-03-23
Fix sorting in image navigator

mdashlw:
### Before you begin

---


MareStare:
# QA

Now the default sorting when no `sf` is specified is deterministic. Before it was using just the `first_seen_at` field:
```
https://derpibooru.org/images/{image_id}?sort[]={image_id}
```
After this change it uses both the `first_seen_at` and the `id` by default in the navigator:
```
http://localhost:8080/images/5?sort[]=1746812956000&sort[]=5
```

PR #475, opened 2025-03-23
More mod logs

mdashlw:
### Before you begin

---


MareStare:
# QA

CRUD for adverts including images is properly logged in mod logs 

image

Unfortunatelly, what exactly the updated diff was is not visible, but I guess it's just the limitation of the current model.

---

Badge creation and update is logged just fine.

However, if the name of the badge is updated, this shows the new name of the badge rather than the old one. @mdashlw WDYT, should we display the old title instead?


Uploading non-svg file as a badge triggers this error, which I think is not specific to this change
image

Other activities work just fine as well

image

What tag change was reverted isn't clear here, but that's at least something

PR #479, opened 2025-03-24
Update home image query for staff users

### LINKED ISSUE #426: Show newest images in Activity
https://github.com/philomena-dev/philomena/blob/7673183d6fbc0e6d45796306c56f5e51dbe663df/lib/philomena_web/controllers/activity_controller.ex#L23-L28
https://github.com/philomena-dev/philomena/blob/7673183d6fbc0e6d45796306c56f5e51dbe663df/lib/philomena_web/controllers/image_controller.ex#L41

Either a setting or just a thing for staff to load latest images and images without thumbnails on the home page

mdashlw:
### Before you begin

---

closes #426


Meow:
Make it a setting staff can toggle on/off, i can see it being useful to see what users see.

liamwhite:
I'm curious why do this if you can use search *?

lib/philomena_web/templates/setting/edit.html.slime
```diff
@@ -109,6 +109,15 @@ h1 Content Settings
           => label f, :scale_large_images
           = error_tag f, :scale_large_images
 
+        - delay_home_images = if staff?(@conn.assigns.current_user), do: :staff_delay_home_images, else: :delay_home_images
+        = field_with_help( \
+          "Delay images on the home page for 3 minutes and until thumbnails are generated",
```
Meow, line 114: "Show images on the homepage with a 3 minute delay, to ensure that the thumbnails are generated, and to allow uploaders to check their tagging after upload."

MareStare: 3 minute delay isn't there to ensure the thumbnails are generated. There is a separate condition that ensures that:
```
must_not: [%{term: %{thumbnails_generated: false}}]
```
Thus, the current text proposed by @mdashlw is more accurate. Could be slightly improved like this:

```suggestion
          "Hide images without thumbnails on home page, and for the first 3 minutes after upload (to allow for last-minute tag updates)",
```

But it probably makes sense to split this into two separate settings - for "no thumbnails generated" and "3 minute delay"

MareStare: Another approach may be just a single input for "Home page query" which defaults to the current `created_at.lte:3 minutes ago, -thumbnails_generated:false`

MareStare:
# QA

Images without thumbnails are shown by default for Administrator, if this setting is enabled:
image

then images without thumbnails are not shown. 

Regular users also have access to this setting, but the delay is enabled by default for them.

Not clear why we need separate `staff_delay_home_images` and `delay_home_images` boolean configs. @mdashlw what's the reason to make two different fields here? Couldn't we just use a single field here?

mdashlw:
@MareStare for different defaults for regular users and staff

mdashlw:
@MareStare feel free to do this if you want.

MareStare:
After I did this change (https://github.com/philomena-dev/philomena/pull/561) I realized that with my suggested approach the following will happen:
- User creates a fresh account
- User goes to its "Display" settings, the default `delay_home_images: true` checkbox is set.
- User changes the values of unrelated settings, and clicks "save"
- The value `true` for `delay_home_images` is now set in the DB column `delay_home_images` instead of `null`
- The user is promoted to a staff user
- The user already has the value `true` in `delay_home_images`, so its `delay_home_images` stays `true` even after they became staff

With separate `staff_delay_home_images` and `delay_home_images` columns the value of delay changes appropriately when the user is promoted/demoted to/from staff automatically. So, better keep it this way

PR #483, opened 2025-03-30
Fix the tag updates atomicity and deadlocks

MareStare:
### Before you begin

---

I've noticed unique constraint violation errors


MareStare:
Found this error during seeding after recent change, trying to debug:

```
app-1  | [error] Task #PID<0.1927.0> started from #PID<0.1926.0> terminating
app-1  | ** (Ecto.NoResultsError) expected at least one result but got none in query:
app-1  | 
app-1  | from i0 in Philomena.Images.Image,
app-1  |   where: i0.id == ^30
app-1  | 
app-1  |     (ecto 3.12.5) lib/ecto/repo/queryable.ex:164: Ecto.Repo.Queryable.one!/3
app-1  |     (philomena 1.1.0) lib/philomena/images/thumbnailer.ex:77: Philomena.Images.Thumbnailer.generate_thumbnails/1
app-1  |     (philomena 1.1.0) lib/philomena/workers/thumbnail_worker.ex:6: Philomena.ThumbnailWorker.perform/1
app-1  |     (exq 0.19.0) lib/exq/worker/server.ex:175: anonymous fn/4 in Exq.Worker.Server.dispatch_work/3
app-1  |     (elixir 1.18.1) lib/task/supervised.ex:101: Task.Supervised.invoke_mfa/2
app-1  | Function: #Function<0.120145007/0 in Exq.Worker.Server.dispatch_work/3>
app-1  |     Args: []
```

MareStare:
Hm, can't reproduce that error mode after `docker compose down`. I suppose I just had my DB messed up where no images where uploaded. Reopenning this again

MareStare:
@Meow I discussed this deadlock in PM with Liam. Note that postgres logs every deadlock occurrence in its sever logs like this:

```
postgres-1  | 2025-03-31 21:18:23.319 UTC [289] DETAIL:  Process 289 waits for ShareLock on transaction 149633; blocked by process 296.
postgres-1  |   Process 296 waits for ShareLock on transaction 149621; blocked by process 289.
postgres-1  |   Process 289: UPDATE "tags" AS t0 SET "images_count" = t0."images_count" + $1 WHERE (t0."id" IN (SELECT st0."id" FROM "tags" AS st0 WHERE (st0."id" = ANY($2)) ORDER BY st0."id" FOR UPDATE))
postgres-1  |   Process 296: UPDATE "tags" AS t0 SET "images_count" = t0."images_count" + $1 WHERE (t0."id" IN (SELECT st0."id" FROM "tags" AS st0 WHERE (st0."id" = ANY($2)) ORDER BY st0."id" FOR UPDATE))
```

There it has good enough context on which queries trapped into a deadlock so it becomes much easier to find the root cause. Postgres never actually hangs on a deadlock, it just detects them and cancels all participating transactions, therefore it's technically not a critical error.

I recommend you to dump the logs from postgres in prod and grep them for `ERROR:  deadlock detected` (not in log snippet higher, but it's also nearby it) to check if there are any real deadlocks happening in prod today.

lib/philomena/images.ex
```diff
@@ -96,22 +96,25 @@ defmodule Philomena.Images do
     |> Multi.insert(:image, image)
     |> Multi.run(:added_tag_count, fn repo, %{image: image} ->
       tag_ids = image.added_tags |> Enum.map(& &1.id)
-      tags = Tag |> where([t], t.id in ^tag_ids)
 
-      {count, nil} = repo.update_all(tags, inc: [images_count: 1])
+      count = Tags.update_images_counts(repo, +1, tag_ids)
```
Meow, line 100: can this not be written simply as `1` instead of `+1`?

MareStare: Removed the `+`

lib/philomena/tags.ex
```diff
@@ -557,6 +626,27 @@ defmodule Philomena.Tags do
     |> Repo.all()
   end
 
+  @doc """
+  Accepts IDs of tags and increments their `images_count` by 1.
+  """
+  @spec update_images_counts(term(), integer(), [integer()]) :: integer()
+  def update_images_counts(repo, diff, tag_ids) do
+    case tag_ids do
```
Meow, line 634: could this `case` statement not be eliminated by using pattern matching in the function definition like `def update_image_counts(repo, diff, tag_ids) when length(tag_ids) == 0, do: 0`

Meow: i'd also rename the function to `update_image_counts`, instead of plural `images`

lib/philomena/tags.ex
```diff
@@ -39,43 +81,72 @@ defmodule Philomena.Tags do
...
-          %Tag{}
-          |> Tag.creation_changeset(%{name: name})
-          |> Repo.insert()
-
-        %{tag | implied_tags: []}
-      end)
+      |> Enum.sort()
+      |> Enum.dedup()
+      |> Enum.map(
+        &(%Tag{}
```
Meow, line 97: i'd not use this shorthand form here for clarity's sake, and instead would write `Enum.map(fn tag_name -> ...`

lib/philomena/tags.ex
```diff
@@ -545,9 +616,7 @@ defmodule Philomena.Tags do
 
         tag_ids = Enum.map(taggings, & &1.tag_id)
 
-        Tag
-        |> where([t], t.id in ^tag_ids)
-        |> Repo.update_all(inc: [images_count: 1])
+        update_images_counts(Repo, +1, tag_ids)
```
Meow: remove the `+` here as well

lib/philomena/tags.ex
```diff
@@ -629,22 +629,20 @@ defmodule Philomena.Tags do
...
-  def update_images_counts(repo, diff, tag_ids) do
-    case tag_ids do
-      [] ->
-        0
+  @spec update_image_counts(term(), integer(), [integer()]) :: integer()
+  def update_image_counts(repo, diff, tag_ids)
 
-      _ ->
-        locked_tags = vectorized_mutation_lock("FOR NO KEY UPDATE", tag_ids)
+  def update_image_counts(nil, _diff, []), do: 0
```
Meow, line 635: match repo to nil? what if repo isn't nil but tag_ids is []?

MareStare: Right, I haven't noticed that `nil`. Copilot autocomplete betrayed me there :skull:

PR #485, opened 2025-04-02
Autocomplete Popup: Fixed `Enter` key presses not hiding popup in the Tag Editor on Android devices

koloml:
### Before you begin

---

This PR fixes an issue I've noticed while trying to edit the tags on my Android smartphone. When I'm typing only part of the tag I need and use autocomplete suggestion to apply the tag, then popup is properly hiding itself. But when I try to type the whole tag without using the popup and then hit "Enter" on my keyboard, then popup just stays visible and out of place.

I was able to reproduce this issue on both Chrome & Firefox browsers using different keyboards I have (Gboard and custom keyboard from my password manager). In both cases popup just stays open until I click out somewhere or start typing the next tag. Recording of the issue on Derpibooru in Chrome:

https://github.com/user-attachments/assets/a44ca348-557d-41b3-8934-e0faee73e6ca

Using DevTools I've captured the keyboard event sent to the script:

image

For some reason, `code` property is empty, while `key` contains correct `Enter` value. I've checked what desktop browsers are doing, and they're both sending `code` and `key` as `Enter`.

After this fix, popup is properly closing itself when Enter is pressed:

https://github.com/user-attachments/assets/dda2cb1d-9863-4fef-ba7b-535e8052cfe5


assets/js/autocomplete/index.ts
```diff
@@ -243,13 +243,21 @@ class Autocomplete {
...
       return;
     }
-    if ((event.key === ',' || event.code === 'Enter') && this.input.type === 'single-tag') {
+
+    let keyCode = event.code;
+
+    // Chrome & Firefox on Android devices return empty code when "Enter" is pressed.
+    if (!keyCode && event.key === 'Enter') {
+      keyCode = event.key;
+    }
```
MareStare, line 252: I think a comma should also confirm the tag in `single-tag` mode, but it doesn't work on my phone. Could you check that as well? Also, what browser are you using to get the dev tools on the phone directly?

koloml: > I think a comma should also confirm the tag in `single-tag` mode, but it doesn't work on my phone. Could you check that as well?

I've checked and it returned `Unidentified` as `key` and empty string as `code`. But it is returning the deprecated `keyCode` property with value of `229`, which is not related to `,` symbol. According to [some document](https://lists.w3.org/Archives/Public/www-dom/2010JulSep/att-0182/keyCode-spec.html) I googled up, `keyCode` of 229 means "Input Method Editor is processing key input".

<details>
  <summary>Screenshot of the event</summary>

image

</details>

I think it's not the first time I encountered this issue, but I can't recall what I did to work around that.

> Also, what browser are you using to get the dev tools on the phone directly?

Kiwi Browser is based on Chrome and includes actual DevTools from the desktop browser. Quite handy.

koloml: Actually, it might be caused by the text prediction features of keyboard. In any case it would be better to just make follow-up PR once we will find the solution to this issue.

assets/js/autocomplete/__tests__/keyboard.spec.ts
```diff
@@ -1,114 +1,146 @@
...
-    }
-  `);
+import { init, TestContext } from './context';
+
+describe('Autocomplete keyboard navigation', () => {
+  let ctx: TestContext;
+
+  beforeAll(async () => {
+    ctx = await init();
+  });
```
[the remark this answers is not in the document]
koloml: I just pushed all the changes I made, so it would be easier to see the issue.

koloml:
Reverted only the last commit. Bumped dependencies seems to be working fine.

PR #489, opened 2025-04-04
Caddy

Meow:
This replaces OpenResty with Caddy as the webserver for the dev environment.

Rationale:
- Written in Go (a highly concurrent and memory-safe language)
- Support plugins via xcaddy out of the box, no need to [run a fork that supports scripting/plugins](https://openresty.org/en/)
- Much more compact and human-readable configs
- Decent/sane default config values

Needs testing/exploring:
- ~~Whether using s3-proxy plugin is actually necessary, as opposed to https://github.com/sagikazarmark/caddy-fs-s3~~
- ~~Whether every response is identical to OpenResty counterpart~~
- ~~Whether caching works / adjust caching settings~~ => caching was deliberately removed, in prod we'll likely have to use a different solution, as caddy caching plugin is extremely slow, unless the devs fix it
- ~~Potentially a way to use multiple different S3 backends/buckets with different credentials~~ => can be accomplished via profiles
- ~~Is there a better way to accomplish what the config does right now?~~


docker/web/config/Caddyfile
```diff
@@ -0,0 +1,50 @@
...
+	cache
+}
+
+(s3path) {
+	route {args[0]} {
+		cache
+		rewrite {args[0]} {args[1]}
+		s3proxy {
+			bucket {$S3_BUCKET}
+			region auto
```
Veetaha, line 11: Is it possible to use spaces for indention? Seems like `.editorconfig` should enforce that

https://github.com/philomena-dev/philomena/blob/2a0a885f34fbed5a6f7138721430752df85f89af/.editorconfig#L1-L9

docker/web/config/Caddyfile
```diff
@@ -0,0 +1,50 @@
+{
+	cache
+}
+
+(s3path) {
+	route {args[0]} {
+		cache
+		rewrite {args[0]} {args[1]}
+		s3proxy {
```
MareStare, line 9: Is it possible to use spaces for indention? Seems like `.editorconfig` should enforce that

https://github.com/philomena-dev/philomena/blob/2a0a885f34fbed5a6f7138721430752df85f89af/.editorconfig#L1-L9

Github renders tabs very inconsistently in different views:
image
image

Meow: Nope, Caddy formatter enforces tabs (`caddy fmt`).

MareStare: I see, let's add the `caddy fmt` check to see

liamwhite: This shouldn't have been resolved because the check wasn't added. Unresolving

Meow: oh crap i misclicked on mobile when trying to view, i was trying to collapse this

MareStare:
Was this closed by mistake? Caddy seems like an overall improvement over openresty

Meow:
> Was this closed by mistake? Caddy seems like an overall improvement over openresty

No. I just don't have it in me to continue working on it, because this change is clearly not wanted, not needed and my work is basically just pointless.

reopen if you want to take over working on this, i can't take this anymore for now

MareStare:
@Meow this change is wanted. I'm sorry to hear you feeling bad, but the April Fools was a success, and I think, in general, user engagement on the site is something we should improve in the long run.

If there are any cheaters/haters, don't let them win. People are very stupid all around and fixing all of them isn't good for mental health. So don't take it too close. Better spend that me tal energy on deflecting that stupidity with code, which feels no emotions.

I hope to see you recover and proceed with this effort, since you have the most competence to get it done right. Remember: the grind never stops 💪🤙

Meow:
It's not about them, or the event. It's about people whose opinion I actually care about just kinda not wanting this.

docker-compose.yml
```diff
@@ -1,6 +1,7 @@
 volumes:
   postgres_data: {}
   opensearch_data: {}
+  caddy_data: {}
```
MareStare, line 4: I haven't worked with nginx/openresty and basically any other kind of pre-made web servers, but I wonder why openresty setup did require this volume before. Or maybe not having that volume for openresty was an oversigh?  

From [caddy docs](https://caddyserver.com/docs/conventions):
> The data directory must not be treated as a cache. Its contents are not ephemeral or merely for the sake of performance. Caddy stores TLS certificates, private keys, OCSP staples, and other necessary information to the data directory. It should not be purged without understanding the implications.

Meow: Resty didn't require it like Caddy does.

Veetaha (approved):

docker/web/config/Caddyfile
```diff
@@ -0,0 +1,72 @@
...
+	import s3path @imgview /images/{re.imgview.1}/{re.imgview.2}/full.{re.imgview.3}
+	import s3path @img /images/{re.img.1}
+	import s3path @avatars /avatars/{re.avatars.1}
+	import s3path @spns /adverts/{re.spns.1}
+	import s3path @badges /badges/{re.badges.1}
+	import s3path @tags /tags/{re.tags.1}
+
+	# This block and the next block after it (the header one) are required
+	# because it is possible for there to be files in the bucket which have
+	# an incorrect content type (e.g. "application/octet-stream").
```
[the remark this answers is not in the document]
Meow: i mean technically anything can be put into a bucket, we don't know the content type, so it's good to force it to have deterministic behavior. We can always fuck up a validation on the uploading server's side... and yes the reason is mostly legacy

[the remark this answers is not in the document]
liamwhite: Incorrect content type is an artifact of the pre-S3 system, and is particularly relevant for s3proxy's filesystem handling which will serve files that were made before it but without a proper content type.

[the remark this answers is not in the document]
Meow: > Incorrect content type is an artifact of the pre-S3 system
incorrect, you can clearly still submit invalid objects to s3

liamwhite (approved):

PR #481, opened 2025-03-27
Add more data to the dev seeds

MareStare:
### Before you begin

- I understand my contributions may be rejected for any reason
- I understand my contributions are for the benefit of Derpibooru and/or the Philomena software
- I understand my contributions are licensed under the GNU AGPLv3

* [x] I understand all of the above

---

This PR cherry-picks two commits from the `redesign` branch (2d5da618aaedc397b14660f99a40153254b1dba5 and 16978b35e2fa1e7a0b7344cc1e0f1b41543237a5).

Note that with all this new seed data the initial startup now takes several minutes (haven't measured exactly), but I think it's worth it.

## Improved logging

Added command logging with color highlight to the dev scripts, so now the logs look like this:

image

I also noticed this warning in the seeding logs:

```

app-1  | [info] Elixir.Philomena.IndexWorker[3a594501-b02a-4ef8-875c-6a4511db8149] done: 24ms
app-1  | identify: unknown image property "%[profile:icc]" @ warning/property.c/InterpretImageProperties/4238.
```

Plus this warning:

https://github.com/philomena-dev/philomena/blob/78550a632272cc46bd0bea4515c265c3acc0aa03/lib/philomena_media/objects.ex#L200

And, there was also the following error:

```
app-1  | [error] Task #PID<0.98808.0> started from #PID<0.98805.0> terminating
app-1  | ** (KeyError) key :body not found in: nil
app-1  |
app-1  | If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
app-1  |     (philomena 1.1.0) lib/philomena_media/objects.ex:72: PhilomenaMedia.Objects.download_file/2
app-1  |     (philomena 1.1.0) lib/philomena/images/thumbnailer.ex:147: Philomena.Images.Thumbnailer.download_image_file/1
app-1  |     (philomena 1.1.0) lib/philomena/images/thumbnailer.ex:78: Philomena.Images.Thumbnailer.generate_thumbnails/1
app-1  |     (philomena 1.1.0) lib/philomena/workers/thumbnail_worker.ex:6: Philomena.ThumbnailWorker.perform/1
app-1  |     (exq 0.19.0) lib/exq/worker/server.ex:175: anonymous fn/4 in Exq.Worker.Server.dispatch_work/3
app-1  |     (elixir 1.18.1) lib/task/supervised.ex:101: Task.Supervised.invoke_mfa/2
[...]

PR #491, opened 2025-04-06
Remove now useless `COPY` instructions because we use a volume bind-mount instead

MareStare:
### Before you begin

---

This PR cherrypicks a commit from https://github.com/philomena-dev/philomena/pull/481


Meow:
Sure, lgtm

PR #492, opened 2025-04-06
Add `philomena` script wrapper over `docker compose`

MareStare:
### Before you begin

---

The script defines a `DROP_DB` env variable that clears all DBs to trigger re-seeding from the fresh state


Meow:
Seems unnecessary. This same stuff could just be documented somewhere.

MareStare:
Why documenting it when it could be scripted?

Meow:
I'd prefer it to ideally just be something that somehow is in the $PATH, so the UX of using it could be similar to [sail](https://github.com/laravel/sail).

MareStare:
I wish there was a way to add repo-local commands automatically into path when one is inside of the repo tree.

I can extend the root-level npm scripts:
https://github.com/philomena-dev/philomena/blob/c679081a3bf199aa0df080686dbd081e28fbba69/package.json#L1-L8

So, that it could be run as `npm run dev`. @Meow WDYT?

Meow:
That's a bit janky, maybe just have a philomena.sh script that automatically sets up the $PATH if user chooses to do so, and then you can simply just use it like
`philomena up`
`philomena down`
`philomena clean`
etc

MareStare:
@Meow I've changed the `dev.sh` to `philomena.sh` plus added a `scripts/path` directory with an extensionless `philomena` wrapper that is intended for adding to `PATH`.

See the updated rendered README.md here: https://github.com/MareStare/philomena/tree/feat/drop-db-script?tab=readme-ov-file#getting-started

Re-requesting your review

README.md
```diff
@@ -4,32 +4,48 @@
...
 ## Getting Started
 
-On systems with `docker` and `docker compose` installed, the process should be as simple as:
+Make sure the following core tools are installed on your machine:
 
-```
-docker compose build
-docker compose up
+- [Docker](https://docs.docker.com/engine/install/) and [Docker Compose plugin](https://docs.docker.com/compose/install/#scenario-two-install-the-docker-compose-plugin)
+- [NodeJS toolchain](https://nodejs.org/en)
```
Meow, line 10: Why is all except docker necessary to be installed on the host? Just run these things via Docker, like instead of starting the app, just run the scripts.

scripts/philomena.sh
```diff
@@ -0,0 +1,90 @@
...
+
+# Clean up everything: DBs, build caches, etc.
+function clean {
+  step docker compose down --volumes
+  step docker container prune --all --force
+  step docker volume prune --all --force
+  step docker image prune --all --force
+  step sudo chown --recursive "$(id -u):$(id -g)" .
+
+  # We don't run a `git clean` here because some people store dirty scripts
```
Meow, line 68: Just prompt whether git cleanup should be ran or not?

Meow: ...or add a flag like `philomena clean --git`

README.md
MareStare: Hm, makes sense

scripts/philomena.sh
MareStare: Makes sense

README.md
```diff
@@ -4,32 +4,48 @@
...
+- [Rust toolchain](https://www.rust-lang.org/tools/install)
+- [Elixir toolchain](https://elixir-lang.org/install.html)
+
+Add the directory `scripts/path` to your `PATH` to get the `philomena` dev CLI globally available in your terminal.
```
Meow, line 14: Add an example how to do that, and preferably something copy-pastable. Might be a good idea to make scripts that are intended to run from `source` block to auto-add to $PATH. Also accomodate both bash and fish shells. zsh too if you're feeling fancy.

MareStare: I think a simple example with 
```
export PATH="$PATH:$HOME/philomena/scripts/path"
```
should be enough. It works for me in `zsh`. Supporting all the various shells is a bit too much I think, people have quite different configs.

I also considered just writing a `philomena` wrapper to `/usr/local/bin` in the init script that would make the setup seamless, but I'm hesitant about that today

Meow: I'd not support more shells than bash, fish and zsh.

Bash is the most popular shell.
Fish is the shell we use in prod deployment.
Zsh is another very popular shell.

I don't care about other shells.

MareStare: I tried to make a file for sourcing but the problem is that for the sourced file the `BASH_SOURCE` appears to be empty
image

At this point, I'm fine with just the path modification example in README. I don't think we'll need anything other than just  the `philomena` command in the shell RC in observable future

MareStare:
Did a rebase, which I guess was unnecessary, but well, it's no going back. The new changes are starting from https://github.com/philomena-dev/philomena/pull/492/commits/0c0376b26e1ee414e48bc4dc338006574f62e6ac

MareStare:
@Meow I propose to merge this at this state. I'll tackle the transition of current `init.sh` script tooling installation to docker separately as part of https://github.com/philomena-dev/philomena/pull/493. Because right now this and that PR start requiring some consolidation

scripts/philomena.sh
MareStare: Removed these requirements from readme. I'll tackle moving stuff to docker in https://github.com/philomena-dev/philomena/pull/493

liamwhite (approved):

MareStare:
Merging this without @Meow's explicit approval just to get it quicker into `master` and iterate on it further in https://github.com/philomena-dev/philomena/pull/493. I believe I have addressed all the review comments so far, and if you'd like to do another round of review, feel free to do so in this PR, I can fix any remaining issues by opening separate PRs.

PR #493, opened 2025-04-06
Add `shellcheck` shell script files linter

MareStare:
### Before you begin

---

See the [tool repo](https://github.com/koalaman/shellcheck) for details. Yeah it's written in Haskell


scripts/install/typos.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+. "$(dirname "${BASH_SOURCE[0]}")/lib.sh"
+
+TOOL_VERSION=${TOOL_VERSION:-1.31.1}
```
MareStare, line 7: Note that we are now pinning the version of `typos`, when previously we were using the `master` version of `typos` on CI. I think pinning the tool version is actually better in general, especially in case of `typos` because `typos` can produce new lints between patch releases. So every time `typos` does a release it breaks Philomena's CI.

Instead, we should bump the version of `typos` explicitly.

Meow:
I... think this is going a tad too far with all the linting, considering you're adding this huge thing and this whole additional step, for the sake of files which we almost never modify.

MareStare:
I can remove it from the precommit hook if it's perf is concerning you, but I expect to extend the dev scripts over time. I don't think this linter costs anything. I specifically made so that adding new linters is extremely easy in this PR by splitting the existing init.sh script into several files per each tool, otherwise the changes in this PR would be much smaller

MareStare:
Superseded by https://github.com/philomena-dev/philomena/pull/528

PR #495, opened 2025-04-07
Add `tracing`'s `EnvFilter`-like log filtering with `PHILOMENA_LOG` env var

MareStare:
### Before you begin

---

I've been debugging way too many things with the https://github.com/philomena-dev/philomena/pull/481, and the lack of the logs has been a really big problem. I haven't found a good way to configure filtering the logs from different modules and third-party libraries dynamically with Elixir's logger, so I rolled my own solution with the `PHILOMENA_LOG` env var.

Also had to add more logs to some places to see what's actually happening there. For example, there was a bug in `s3-proxy` with uploading large files, but the error wasn't properly logged, now it is:

image
image

You can enable debug logging in the `Objects` module like this:
```
PHILOMENA_LOG-Ecto=none,Exq=none,PhilomenaMedia.Objects=debug,debug
```
And then see these logs:
image

As for the thumbnail generation logs - without them I would never know that my `valkey` job queue was completely full with a bunch of duplicate thumbnail generation jobs, and that I needed to clean it up manually.

> On an unrelated note. I had experience with implementing a job queue at work, and we used a job `fingerprint` mechanism that was designed to solve this exact problem - if there is already a job in the queue with the same fingerprint, then a new one is not queued to avoid duplicate work. The fingerprint is derived from what makes the job's "identity" - i.e. its kind and parameters.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,7 @@ services:
       # event's module and function. The last entry in the list of filters should
       # be a bare `level` which will be used as a catch-all for all other log
       # events that do not match any of the previous filters.
-      - PHILOMENA_LOG=${PHILOMENA_LOG-Ecto=none,Exq=none,PhilomenaMedia.Objects=info,debug}
+      - PHILOMENA_LOG=${PHILOMENA_LOG-Ecto=debug,Exq=none,PhilomenaMedia.Objects=info,debug}
       - MIX_ENV=dev
       - PGPASSWORD=postgres
       - ANONYMOUS_NAME_SALT=2fmJRo0OgMFe65kyAJBxPT0QtkVes/jnKDdtP21fexsRqiw8TlSY7yO+uFyMZycp
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,12 +14,11 @@ services:
       dockerfile: ./docker/app/Dockerfile
     environment:
       # Use this env variable to control the logs levels from different modules.
-      # The syntax is `{module.function}=level,...` and the `module.function`.
-      # The filter evaluation stops on the first `{module.function}` that has a
-      # prefix match against the log event's module and function. The last entry
-      # in the list of filters should be a bare `level` which will be used as a
-      # catch-all for all other log events that do not match any of the previous
-      # filters.
+      # The syntax is `{module.function}=level,...`. The filter evaluation stops
+      # on the first `{module.function}` that has a prefix match against the log
+      # event's module and function. The last entry in the list of filters should
+      # be a bare `level` which will be used as a catch-all for all other log
+      # events that do not match any of the previous filters.
       - PHILOMENA_LOG=${PHILOMENA_LOG-Ecto=none,Exq=none,PhilomenaMedia.Objects=info,debug}
       - MIX_ENV=dev
       - PGPASSWORD=postgres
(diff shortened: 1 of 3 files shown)


lib/philomena/application.ex
```diff
@@ -56,4 +58,58 @@ defmodule Philomena.Application do
...
+                Logger.compare_levels(event.level, level) != :lt
+            end
+        end
+      end
+
+      :logger.add_primary_filter(
+        :sql_logs,
+        {fn event, _ -> if(allow_log_event?.(event), do: :ignore, else: :stop) end, []}
+      )
+    end
```
MareStare, line 113: Is there a better way to write this code? The lack of early returns in Elixir is really a bummer. What I effectively want to it to look like is this to avoid the code nesting problem:

```ex
    if Enum.empty?(filters) do
      return
    end

    allow_log_event? = fn event ->
      mfa = Map.get(event.meta, :mfa)

      if mfa == nil do
        return false
      end

      {module, function, _arity} = mfa
      scope = "#{inspect(module)}.#{function}"

      filter = filters |> Enum.find(fn {selector, _level} ->
        is_nil(selector) or String.starts_with?(scope, selector)
      end)

      if is_nil(filter) do
        return false
      end
      

      {_selector, level} = filter
      
      Logger.compare_levels(event.level, level) != :lt
    end

    :logger.add_primary_filter(
      :sql_logs,
      {fn event, _ -> if(allow_log_event?.(event), do: :ignore, else: :stop) end, []}
    )
  end
```

MareStare: Liam suggested `with` operator in DM, will clean this up

docker-compose.yml
```diff
@@ -13,6 +13,13 @@ services:
       context: .
       dockerfile: ./docker/app/Dockerfile
     environment:
+      # Use this env variable to control the logs levels from different modules.
+      # The syntax is `{module.function}=level,...`. The filter evaluation stops
+      # on the first `{module.function}` that has a prefix match against the log
+      # event's module and function. The last entry in the list of filters should
+      # be a bare `level` which will be used as a catch-all for all other log
+      # events that do not match any of the previous filters.
+      - PHILOMENA_LOG=${PHILOMENA_LOG-Ecto=none,Exq=none,PhilomenaMedia.Objects=info,debug}
```
Meow, line 22: why not log ecto? i often resort to seeing the exact SQL executed to see where i screwed up

MareStare: By default - SQL logging is just extremely noisy, but anyway I can keep it on by default if you insist.

To explain my thinking for the long term:

My reason to disable it by default is that I found them extremely noisy, but anyway I can't do anything with that in the terminal. Terminal is too limited for working with logs at large.

My long-term wish is to send logs to Loki and view them in Grafana where it's easy to have a brids-eye view on the general state of the app via a dashboard like this. (this is a screenshot from my [Telegram bot](https://github.com/Veetaha/snowpity) that uses Grafana, Loki, VictoriaMetrics):

image

From this view you can quickly see the general metrics, and the last panic / error / warning logs.

Then you can drill down into the gory details of the logs using the LogQL syntax 
image

Then expand the details of every log event you are interested in 
image

This way there is no noise by default, it's convenient to view logs, and debug any request.

MareStare: I've rebased this branch since it's been rather old. I pushed a single commit new change [a94c676](https://github.com/philomena-dev/philomena/pull/495/commits/a94c6762b68a3a7d30d5588fe6cfcd3733984cbb) to set ecto logging level to `debug`. Everything else is the same

PR #505, opened 2025-04-16
Image approval checks

mdashlw:
### Before you begin

---


Meow:
You should be able to approve deleted images. Also, the error text is wrong either ways, should say "cannot approve a deleted image" just like it does in the rest of the code base.

PR #506, opened 2025-04-16
Always render the search results page even if no images

### LINKED ISSUE #496: Display "Show hidden/deleted" buttons on a "No images found" page

mdashlw:
### Before you begin

---

closes #496


lib/philomena_web/views/search_view.ex
```diff
@@ -3,11 +3,4 @@ defmodule PhilomenaWeb.SearchView do
 
   def scope(conn), do: PhilomenaWeb.ImageScope.scope(conn)
   def hides_images?(conn), do: can?(conn, :hide, %Philomena.Images.Image{})
-
-  def override_display(conn, [{tag, _description, dnp_entries}]) do
-    tag.images_count > 0 or Enum.any?(dnp_entries) or
-      (present?(tag.mod_notes) and can?(conn, :edit, tag))
```
MareStare, line 9: Do you know what was all this complicated logic about? Seems rather convoluted given that it only influenced wether you saw the empty image view or not

MareStare (approved):
Tested this locally. Seems okay to merge.

Before:
image

After:
image

lib/philomena_web/views/search_view.ex
mdashlw: It became complicated only recently, used to be just `any?(images)`. I don't think it was ever worth it

PR #507, opened 2025-04-16
Clean tags job

mdashlw:
### Before you begin

---

Run daily/weekly on production. First run on Derpibooru will nuke 120k tags

### DIFF
--- a/docker/app/run-development
+++ b/docker/app/run-development
@@ -27,6 +27,7 @@ background() {
     step mix run -e 'Philomena.Release.update_stats()' > /dev/null
     step mix run -e 'Philomena.Release.clean_moderation_logs()' > /dev/null
     step mix run -e 'Philomena.Release.generate_autocomplete()' > /dev/null
+    step mix run -e 'Philomena.Release.clean_tags()' > /dev/null
 
     sleep 300
   done
--- a/lib/philomena/tags.ex
+++ b/lib/philomena/tags.ex
@@ -732,4 +732,55 @@ defmodule Philomena.Tags do
   def change_implication(%Implication{} = implication) do
     Implication.changeset(implication, %{})
   end
+
+  @doc """
+  Deletes all tags that meet all of the following conditions:
+  - No category assigned
+  - Zero images count
+  - No description
+  - No short description
+  - No mod notes
+  - No image attached
+  - Not aliased to another tag
+  - Has no aliases pointing to it
+  - Does not imply any other tags
+  - Is not implied by any other tags
+
+  Also removes the deleted tags from the search index.
+
+  ## Examples
+
+      iex> cleanup!()
+      {3, [1, 2, 3]}
+
+  """
+  def cleanup! do
+    tag_ids =
+      from(t in Tag,
+        where: is_nil(t.category),
+        where: t.images_count == 0,
+        where: t.description == "",
+        where: is_nil(t.short_description) or t.short_description == "",
+        where: is_nil(t.mod_notes) or t.mod_notes == "",
+        where: is_nil(t.image),
+        where: is_nil(t.aliased_tag_id),
+        left_join: a in assoc(t, :aliases),
+        left_join: it in assoc(t, :implied_tags),
+        left_join: ibt in assoc(t, :implied_by_tags),
+        where: is_nil(a.id) and is_nil(it.id) and is_nil(ibt.id),
+        select: t.id
+      )
+      |> Repo.all()
+
+    {count, _} =
+      Tag
+      |> where([t], t.id in ^tag_ids)
+      |> Repo.delete_all()
+
+    if count > 0 do
+      PhilomenaQuery.Search.delete_documents(tag_ids, Tag)
+    end
+
+    {count, tag_ids}
+  end
 end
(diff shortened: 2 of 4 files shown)


lib/philomena/tags.ex
```diff
@@ -732,4 +732,55 @@ defmodule Philomena.Tags do
...
+      |> Repo.all()
+
+    {count, _} =
+      Tag
+      |> where([t], t.id in ^tag_ids)
+      |> Repo.delete_all()
+
+    if count > 0 do
+      PhilomenaQuery.Search.delete_documents(tag_ids, Tag)
+    end
```
MareStare, line 782: There may be issues due to the non-atomic delete between Postgres and OpenSearch. For example, what if OpenSearch is not available for some reason - we would then delete the tags from Postgres but then we would failed to do that in OpenSearch and the tags will be left there internally introducing a resynchronization.

There are several ways to go about this problem.

## Two-Step Deletion

1. Mark all tags, that pass the criteria for deletion in Postgres via a soft-delete operation, which can be represented by setting setting `deleted = true` (a new `deleted` column needs to be added) or `delete_marker = { "deletion_date": "...", "deletion_reason": "..." }` if you want to go fancy and make it easier to diagnose faulty deletions.
2. All existing queries need to be updated to exclude all tags with `deleted = true` or `delete_markr IS NOT NULL` which effectively already removes the tag from the system.
3. The same job should then do a separate second query to list all existing soft-deleted tags in Postgres, which will also include tags that this job previously failed to delete from OpenSearch, and then try deleting them in OpenSearch. This will ensure we retry the deletion of previously soft-deleted that we failed to delete in OpenSearch.
4. Now finally hard-delete in Postgres all tags that we successfully removed from OpenSearch.

## Resync / Garbage Collection

Introduce a new job that scans Postgres, and scans OpenSearch to find the differences between them and fix them up, while reporting all the out-of-ordinary inconsistencies like "a tag present in Postgres, but not in OpenSearch for too long", "a tag present in OpenSearch but not in Postgres".

The resync job of this kind will cover all the potential sync issues and bugs.

mdashlw: byte wanted to do a proper indexing job system here https://github.com/philomena-dev/philomena/tree/persistent-indexing but it looks dead now. it's probably fine to assume OpenSearch is available since that's what everything else does

MareStare: The simplest thing we could do here is wrap the entire operation in a Postgres transaction - so that if Opensearch fails, the transaction is cancelled and no data is deleted at all.

MareStare: Keep in mind that requests to OpenSearch can fail not just because of infra outage, but because of general bugs in code anywhere - including the code that forms the request in Philomena and in bugs in OpenSearch internally.

For example, a potential bug could be that the request size exceeds the maximum size allowed by OpenSearch, but usually it's something much less obivous

lib/philomena/tags.ex
```diff
@@ -732,4 +732,55 @@ defmodule Philomena.Tags do
...
+
+      iex> cleanup!()
+      {3, [1, 2, 3]}
+
+  """
+  def cleanup! do
+    tag_ids =
+      from(t in Tag,
+        where: is_nil(t.category),
+        where: t.images_count == 0,
```
Meow, line 761: `images_count` field is unreliable, you should query `select count(*) from image_taggings where tag_id = TAG_ID;`. `images_count` will be set to 0 even if the tag IS actually present on some deleted image, which might leave that deleted image with less than 3 tags. This should never happen.

lib/philomena/tags.ex
```diff
@@ -732,4 +732,55 @@ defmodule Philomena.Tags do
...
+        where: t.description == "",
+        where: is_nil(t.short_description) or t.short_description == "",
+        where: is_nil(t.mod_notes) or t.mod_notes == "",
+        where: is_nil(t.image),
+        where: is_nil(t.aliased_tag_id),
+        left_join: a in assoc(t, :aliases),
+        left_join: it in assoc(t, :implied_tags),
+        left_join: ibt in assoc(t, :implied_by_tags),
+        where: is_nil(a.id) and is_nil(it.id) and is_nil(ibt.id),
```
Meow, line 770: These fields are `has_many` type relations, and thus would be lists...

Meow (changes requested):
I'd also like to see some sort of backup mechanism before this job is ran (perhaps trigger this job manually in production after backup is created, rather than on a set schedule?)

docker/app/run-development
```diff
@@ -27,6 +27,7 @@ background() {
     step mix run -e 'Philomena.Release.update_stats()' > /dev/null
     step mix run -e 'Philomena.Release.clean_moderation_logs()' > /dev/null
     step mix run -e 'Philomena.Release.generate_autocomplete()' > /dev/null
+    step mix run -e 'Philomena.Release.clean_tags()' > /dev/null
```
Meow, line 30: does this really need to run in dev?

lib/philomena/tags.ex
Meow: i'd like to see this wrapped in a transaction, as per @MareStare

Meow (changes requested):

docker/app/run-development
MareStare: Why not if it's part of our app - this way we have a test during development that this job works and does just fail horribly

Meow: i mean yea, but what if i want empty tags for a bit?

PR #515, opened 2025-04-24
Bump Caddy version to 2.10.0

delly-bud:
### Before you begin

---

Caddy S3 FS module requires higher go version, failing the build otherwise.

### DIFF
--- a/docker/web/Dockerfile
+++ b/docker/web/Dockerfile
@@ -1,4 +1,4 @@
-ARG CADDY_VERSION="2.9.1"
+ARG CADDY_VERSION="2.10.0"
 
 FROM caddy:$CADDY_VERSION-builder-alpine AS builder
 
--- a/docker/web/Dockerfile
+++ b/docker/web/Dockerfile
@@ -1,9 +1,10 @@
 ARG CADDY_VERSION="2.10.0"
+ARG CADDY_FS_S3_VERSION=v0.10.0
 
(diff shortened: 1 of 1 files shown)


MareStare (approved):

MareStare:
@delly-bud maybe we should pin the s3 plugin version too to avoid these incompatibility issues?

delly-bud:
@MareStare Fair point. 

Pinned the S3 FS Module version to a specific version, so the compilation issue does not pop up again.

docker/web/Dockerfile
```diff
@@ -1,9 +1,10 @@
-ARG CADDY_VERSION="2.9.1"
+ARG CADDY_VERSION="2.10.0"
+ARG S3_CADDY_VERSION="0.10.0"
 
 FROM caddy:$CADDY_VERSION-builder-alpine AS builder
 
 RUN xcaddy build \
-    --with github.com/sagikazarmark/caddy-fs-s3
+    --with github.com/sagikazarmark/caddy-fs-s3@v0.10.0
```
MareStare, line 7: Looks like you intended to use the `S3_CADDY_VERSION` variable here

```suggestion
    --with github.com/sagikazarmark/caddy-fs-s3@$S3_CADDY_VERSION
```

MareStare (changes requested):
I think some commits from Tantabus were cherry-picked by mistake. Let's remove them

Veetaha (approved):

PR #522, opened 2025-04-25
Fix a couple _id index types

mdashlw:
### Before you begin

---

Will need to rebuild image and report indices


liamwhite (approved):

Meow:
As this change has something to do with indices, this is blocked similarly to #526 (see comments on that PR)

PR #528, opened 2025-04-27
Add devcontainer setup

MareStare:
### Before you begin

---

This PR adds the `.devcontainer` definition allowing us to use a consistent dev environment with predefined configurations and dependencies. Note, however, that running your IDE in a devcontainer is not required. If you'd still like to develop from your host the scripts will spin up a background devcontainer for you (which just sleeps) and forward their execution to that container via `docker exec`, so things should still just work.

This took a lot of time to assemble, I think I've tried almost every configuration design possible at this point and landed on this one.


Meow:
Giant PR, I'm pushing it back to release 1.3.0 as opposed to 1.2.0, as I'll need a significant amount of time to understand it and review it properly.

MareStare:
Superseded by https://github.com/philomena-dev/philomena/pull/644

PR #531, opened 2025-04-28
Link to true_uploader for mods

mdashlw:
### Before you begin

---

### DIFF
--- a/lib/philomena_web/templates/profile/show.html.slime
+++ b/lib/philomena_web/templates/profile/show.html.slime
@@ -28,7 +28,11 @@
         li = link("Report this user", to: ~p"/profiles/#{@user}/reports/new")
 
       ul.profile-top__options__column
-        li = link("Uploads", to: ~p"/search?#{[q: "uploader_id:#{@user.id}"]}")
+        = case @conn.assigns.current_user do
+          - %{role: role} when role in ~W(moderator admin) ->
+            li = link("Uploads", to: ~p"/search?#{[q: "true_uploader_id:#{@user.id}"]}")
+          - _ ->
+            li = link("Uploads", to: ~p"/search?#{[q: "uploader_id:#{@user.id}"]}")
         li = link("Comments", to: ~p"/comments?#{[cq: "user_id:#{@user.id}"]}")
         li = link("Posts", to: ~p"/posts?#{[pq: "user_id:#{@user.id}"]}")
         = if current?(@user, @conn.assigns.current_user) do


Meow:
afaik it's `true_uploader:username`, the ID doesn't work, or at least it didn't a while back...

also yay for bringing back booru on rails behavior

mdashlw:
> afaik it's `true_uploader:username`, the ID doesn't work, or at least it didn't a while back...
> 
> also yay for bringing back booru on rails behavior

both work

MareStare:
#  QA

For `Pleb` and unauthenticated user the `Uploads` links to a search by `uploader_id`
image

For `Administrator` it links to `true_uploader_id` search:
http://localhost:8080/search?q=true_uploader_id%3A4

The problem is that `Recent Uploads` wasn't updated. It still links to a search by `uploader_id`
image

In fact this duplication of UI functionality is really confusing. Do we really need that `Uploads` small link when we already have the `Recent Uploads` with the display of the latest 4 uploads linking to the same search.

Meow:
> # QA
> For `Pleb` and unauthenticated user the `Uploads` links to a search by `uploader_id` image
> 
> For `Administrator` it links to `true_uploader_id` search: http://localhost:8080/search?q=true_uploader_id%3A4
> 
> The problem is that `Recent Uploads` wasn't updated. It still links to a search by `uploader_id` image
> 
> In fact this duplication of UI functionality is confusing. Do we really need that `Uploads` small link when we already have the `Recent Uploads` with the display of the latest 4 uploads plus linking to the same search.

that's ok, that's desired

MareStare:
@Meow could you elaborate why this difference is desired?

Meow:
> @Meow could you elaborate why this difference is desired?

I believe the goal of this PR was to replicate the Booru-on-Rails behavior. This is exactly how it used to be. Thus, it is desired.

MareStare:
@Meow don't you find this old behavior confusing?

mdashlw:
> In fact this duplication of UI functionality is confusing. Do we really need that `Uploads` small link when we already have the `Recent Uploads` with the display of the latest 4 uploads plus linking to the same search.

Yes because it doesn't show recent uploads if all uploads are deleted or anon

MareStare:
> Yes because it doesn't show recent uploads if all uploads are deleted or anon

I see, so maybe we need to show them for mods then in recent uploads? Display the "Recent Uploads" that include deleted/anon uploads if the person viewing is a mod/admin without admin tools turned off.

Because I see no reason why we would need to `Uploads` links on UI. For regular users - it doesn't make sense for sure, because regular users never see deleted or anon images.

mdashlw:
> > Yes because it doesn't show recent uploads if all uploads are deleted or anon
> 
> I see, so maybe we need to show them for mods then in recent uploads? Display the "Recent Uploads" that include deleted/anon uploads if the person viewing is a mod/admin without admin tools turned off.
> 
> Because I see no reason why we would need two `Uploads` links on UI at least for regular users. For regular users - it doesn't make sense for sure, because regular users never see deleted or anon images.

I think it's fine as is.

MareStare:
@mdashlw @Meow  keep in mind that you've been using this UI for so long. I'm very new to it, so I have the valuable perspective of a common users who knows nothing about the system. If I think something is confusing it means it's likely that other 90% of people who are as unfamiliar with this UI and behaviors will think the same. So I stay with my position declaring that this UI design is confusing and I think the majority of users find it too.

It's not a big problem today because  that "Uploads" link is really small and I never even noticed it before I saw this PR. I always go to "Recent Uploads" instead if I'm seeking for uploads.

We can keep this inconsistency for now. But in v2 UI I'm going to be much more pushy for a better and less confusing design. Accepted

PR #538, opened 2025-04-29
Fix scope for artist link page

### LINKED ISSUE #537: Artist link search does not preserve search query
1. Go to artist link requests
2. Search for something
3. Go to page 2
4. Observe that the search query is reset to nothing

mdashlw:
### Before you begin

---

fixes #537

### DIFF
--- a/lib/philomena_web/views/admin/artist_link_view.ex
+++ b/lib/philomena_web/views/admin/artist_link_view.ex
@@ -21,10 +21,17 @@ defmodule PhilomenaWeb.Admin.ArtistLinkView do
     |> String.capitalize()
   end
 
-  def link_scope(conn) do
-    case conn.params["all"] do
-      nil -> []
-      _val -> [all: true]
+  def scope(conn) do
+    []
+    |> scope(conn, "lq", :lq)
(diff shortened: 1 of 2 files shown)


lib/philomena_web/views/admin/artist_link_view.ex
```diff
@@ -21,10 +21,17 @@ defmodule PhilomenaWeb.Admin.ArtistLinkView do
...
+    []
+    |> scope(conn, "lq", :lq)
+    |> scope(conn, "all", :all)
+  end
+
+  defp scope(list, conn, key, key_atom) do
+    case conn.params[key] do
+      nil -> list
+      "" -> list
+      val -> [{key_atom, val} | list]
```
Meow, line 34: Technically not the same behavior as previously. Your code sets the value to `conn.params[key]`, the old code set it to `true`.

mdashlw: It doesn't matter. The controller only checks for presence, not "true"

MareStare:
# QA

Now when you click on the next page of artist links search results, the query is preserved in query params
image

PR #544, opened 2025-05-01
Update UserStatistics.inc_stat to accept user_id

mdashlw:
### Before you begin

---

Accept id instead of object to avoid preloading

Fixes approving a comment incrementing `comments_posted` twice

### DIFF
--- a/lib/philomena_web/controllers/image/comment/approve_controller.ex
+++ b/lib/philomena_web/controllers/image/comment/approve_controller.ex
@@ -18,7 +18,7 @@ defmodule PhilomenaWeb.Image.Comment.ApproveController do
 
     {:ok, _comment} = Comments.approve_comment(comment, conn.assigns.current_user)
 
-    UserStatistics.inc_stat(comment.user, :comments_posted)
+    UserStatistics.inc_stat(comment.user.id, :comments_posted)
 
     conn
     |> put_flash(:info, "Comment has been approved.")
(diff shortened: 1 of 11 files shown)


MareStare (commented):
From the description I expected to see some preloads eliminated, but I don't see any.
So I don't see the problem this solves. Could you point it out?

mdashlw:
> From the description I expected to see some preloads eliminated, but I don't see any. So I don't see the problem this solves. Could you point it out?

#542 has a couple user preloads.

lib/philomena_web/controllers/image/comment/approve_controller.ex
```diff
@@ -3,23 +3,19 @@ defmodule PhilomenaWeb.Image.Comment.ApproveController do
...
-    persisted: true,
-    preload: [:user]
+    persisted: true
 
   def create(conn, _params) do
     comment = conn.assigns.comment
 
     {:ok, _comment} = Comments.approve_comment(comment, conn.assigns.current_user)
 
-    UserStatistics.inc_stat(comment.user, :comments_posted)
```
MareStare, line 21: @mdashlw are you sure this is correct. The comments statistic is not increased unless the comment is approved. So this increment here was looking fine.

Frankly, it's not very clear how comment approvals work. If I post a comment as an anon it always is visible and doesn't require any approval. What is the scenario where a comment is not auto-approved?

mdashlw: It's done here https://github.com/philomena-dev/philomena/blob/eae89d4f93b49fd93541f45a15a314d5cc2b49ca/lib/philomena/comments.ex#L224

MareStare: Ah I see, somehow missed it.

PR #549, opened 2025-05-01
Replace css `has()` with manual state management

MareStare:
### Before you begin

---

The CSS `has()` selector is too new (2023 baseline), so changed the `tags.css` impl to add/remove classes `tag--{state}` for the tag instead

### DIFF
--- a/assets/css/views/tags.css
+++ b/assets/css/views/tags.css
@@ -201,14 +201,14 @@
   margin-left: 5px;
 }
 
-.tag:has(.tag__state[title="Spoilered"]:not(.hidden)),
-.tag:has(.tag__state[title="Hidden"]:not(.hidden)) {
+.tag--spoilered,
+.tag--hidden {
   opacity: 0.6;
 }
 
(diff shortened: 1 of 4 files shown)


MareStare:
Added a PR into this branch to replace the css `has()` here, @Meow please check: https://github.com/philomena-dev/philomena/pull/549

Meow:
Is it possible to do this without object oriented programming? The rest of our JS doesn't really use that.

MareStare:
Why not? Using a class to group related data and get method syntax doesn't mean full it's full oop mode. I can't go without classes, just like I can't imagine Rust without structs.Take a look at `suggestions-view.ts`, `suggestions-model.ts`, and any other code in `autocomplete`.

Meow:
I imagined this change would be a straightforward edit of adding 1 function that sets the state and which also adds the class to whatever it sets the state of... it just seems extremely unnecessary to use classes here

MareStare:
My principle is to leave code in a better state than it was, which means refactoring. If you'd like to keep the original structure, I'd leave this change to you

PR #550, opened 2025-05-02
Approve deleted images

mdashlw:
### Before you begin

* [ ] I understand all of the above

---

### DIFF
--- a/lib/philomena/images.ex
+++ b/lib/philomena/images.ex
@@ -222,7 +222,6 @@ defmodule Philomena.Images do
   def count_pending_approvals(user) do
     if Canada.Can.can?(user, :approve, %Image{}) do
       Image
-      |> where(hidden_from_users: false)
       |> where(approved: false)
       |> Repo.aggregate(:count)
     else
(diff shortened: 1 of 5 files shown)


MareStare (approved):
How does this difference of "deleted but not approved" and "deleted and approved" images manifest in the UI/behaviors?

mdashlw:
> How does this difference of "deleted but not approved" and "deleted and approved" images manifest in the UI/behaviors?

Unapproved images are currently forcibly filtered out from all search queries, even for staff. The request was to change that. Latest idea is to approve on delete since the purpose of the approval system is to temporarily hide images from the public view and it shouldn't have an effect beyond the initial deciding action on the image.


This does create an inconsistent behavior where `uploads_count` Is not incremented and a verification system report is not created but it might be better to address this in the cleanup counters PR.

MareStare:
I see, then it looks like approval and deletion should be mutually exclusive states. I.e. they shouldn't be two different field, but a single field `state: "pending_approval" | "approved" | "hidden_from_users"` if not in DB, then at least at app level. Not to say that we need to change this in this PR, but something to keep in mind and consider in the future

MareStare:
# QA
## Before apply the migration
Deleted but not approved image still show in the approval queue (had to create it on the version of code before this change)
image

After this change, if I delete the image that isn't yet approved, it no longer appears in the approval queue:
image

The image is shown when `+D` is selected
image

## After applying the migration

I noticed it a bit late, but I didn't run the migration. After the migration the approval queue is now empty:

image

However, the old deleted but unapproved images don't appear in searches.

This change https://github.com/philomena-dev/philomena/labels/requires-reindex to make the old images that were deleted but not approved work.

PR #556, opened 2025-05-04
Fix global index link

mdashlw:
quick fix until we have a better solution


MareStare:
# QA

Redirect's working correctly

image

PR #557, opened 2025-05-04
Add link to image on related page

mdashlw:
Similar to other pages like tag changes
image

### DIFF
--- a/lib/philomena_web/templates/image/related/index.html.slime
+++ b/lib/philomena_web/templates/image/related/index.html.slime
@@ -2,9 +2,10 @@
   .block__header
     .block__header__title.hide-mobile
       ' Viewing related images for
-      strong>
-        | #
-        = @image.id
+      = link to: ~p"/images/#{@image}" do
+          strong>
+          | image #
+          = @image.id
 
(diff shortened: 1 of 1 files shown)


lib/philomena_web/templates/image/related/index.html.slime
```diff
@@ -2,9 +2,10 @@
   .block__header
     .block__header__title.hide-mobile
       ' Viewing related images for
-      strong>
-        | #
-        = @image.id
+      = link to: ~p"/images/#{@image}" do
+          strong>
+          | image #
```
Meow, line 7: wrong indentation, needs to be 1 level  higher. Right now this would render as
```
<a href="..."><strong> </strong>image #1234</a>
```

Meow (changes requested):

lib/philomena_web/templates/image/related/index.html.slime
mdashlw: Header links are already bold so removed it

MareStare:
# QA

Link's working fine. Easy accept

image

PR #561, opened 2025-05-05
Refactor separate `delay_home_images`/`staff_delay_home_images` to a single nullable DB column before it's in prod

MareStare:
### Before you begin

---

This is a refactoring as was suggested in https://github.com/philomena-dev/philomena/pull/479#issuecomment-2849692653. The gist is that we make the column nullable and handle the defaulting logic in application code instead.

I added a new migration that removes the two columns that we added in the migration created in https://github.com/philomena-dev/philomena/pull/479, and it re-adds the `delay_home_images` column but as a nullable boolean column with `null` as an implicit default.

I didn't go for re-writing the old migration because updating already applied migrations is a PITA in case if you want to keep your dev env working, so I just added a new migration. This change must be merged before https://github.com/philomena-dev/philomena/pull/479 is applied in prod (cc @Meow)

### DIFF
--- a/lib/philomena/users/user.ex
+++ b/lib/philomena/users/user.ex
@@ -93,8 +93,9 @@ defmodule Philomena.Users.User do
     field :show_hidden_items, :boolean, default: false
     field :hide_vote_counts, :boolean, default: false
     field :hide_advertisements, :boolean, default: false
-    field :delay_home_images, :boolean, default: true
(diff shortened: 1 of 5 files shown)


MareStare:
Closed as per the flaw explained in https://github.com/philomena-dev/philomena/pull/479#issuecomment-2849762371

PR #565, opened 2025-05-09
resize media in media lists without js

Meow:
stolen from the redesign branch

eliminates resizablemedia.js

uses grid for media lists


MareStare:
# QA

Reported in Discord DM, but copying it here to make sure we don't forget to address this.

The featured image with landscape aspect ratio is clipping:

image

On master it looks like this:

image

MareStare:
# QA

It looks like images have some max size that prevents them from growing at some point with the new grid layout. It's as if there is some `max-width` set for the images somewhere.
See the difference between master and this branch below. 

https://github.com/user-attachments/assets/d5853302-f0fb-4c5a-9e4b-26c7f9c292e2

I think the behaviour of scaling on master is fine and should be preserved. It looks weird when the `.media-box` around the image grows, but image stays of the same max size

Meow:
> # QA
> Reported in Discord DM, but copying it here to make sure we don't forget to address this.
> 
> The featured image with landscape aspect ratio is clipping

resolved

> # QA
> It looks like images have some max size that prevents them from growing at some point with the new grid layout. It's as if there is some `max-width` set for the images that's lower than on master. See the difference between master and this branch below.
> 
>  master-branch-grid.mp4 
> I think the behaviour of scaling on master is fine and should be preserved. It looks weird when the `.media-box` around the image grows, but image stays of the same max size

not really a bug? i made it centered though. this "bug" is caused by physical image files being merely 250px on largest side, while the container can be larger than 250px. Old Javascript code ensured the maximum width of the thumb container was 250px and it would never be wider than that. To be honest, images behaving this was is the intended behavior for images whose size is smaller than thumb box (think art that's by itself smaller than 250px). If user has "serve HiDPI thumbnails" option enabled, this issue does not really show up.

assets/css/elements/media.css
```diff
@@ -1,10 +1,39 @@
...
+  grid-gap: var(--padding-small);
+}
+
+.media-list:before {
+  content: "";
+  width: 0;
+  padding-bottom: calc(100% + var(--media-box-header-height));
+  grid-row: 1 / 1;
+  grid-column: 1 / 1;
+}
```
MareStare, line 14: I believe this `:before` plus the `:first-child` on line 34 are not need and can be removed. These are legacy code that doesn't make sense to pull into new code today.

If you do want to keep it, then please add some comments with the explanation, because literally the next person to read/refactor this code will not understand any single bit of it. Code is not kept because "it's sacred" - it's kept only because it solves some problem. If there is no problem to solve - no code is needed. If no reason can be remembered for code to exist - I'd prefer to remove that code and be reminded of it only if someone reports a bug, at which point it will become clear what problem there is, and we may re-view it in a new context again.

MareStare:
# QA

There is some weird layout on the galleries page

image
image

MareStare (commented):
See the last QA item about galleries, otherwise it works, and that's that main thing. Better move fast

MareStare (approved):

PR #567, opened 2025-05-09
Use the actual full resolution 23MB GIF in dev seeds

MareStare:
### Before you begin

---

### DIFF
--- a/priv/repo/seeds_development.json
+++ b/priv/repo/seeds_development.json
@@ -21,7 +21,7 @@
   ],
   "remote_images": [
     {
-      "url": "https://derpicdn.net/img/2015/9/26/988000/thumb.gif",
+      "url": "https://derpicdn.net/img/download/2015/9/26/988000.gif",
       "sources": ["https://derpibooru.org/988000"],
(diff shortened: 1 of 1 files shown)


Meow:
the reason thumb was used was because it doesn't really matter too much for test environment, and download 23MB webm is slow on bad connections, which many people still have. When I was working on Philomena in 2019-2022, I was on 16mbit/sec connection because no ISP wanted to provide a better option for my address because it was "low traffic", and I only got GPON in 2022 after the city ran a municipal project to equip all the households stuck in that situation with fiber.

mdashlw:
> the reason thumb was used was because it doesn't really matter too much for test environment, and download 23MB webm is slow on bad connections, which many people still have. When I was working on Philomena in 2019-2022, I was on 16mbit/sec connection because no ISP wanted to provide a better option for my address because it was "low traffic", and I only got GPON in 2022 after the city ran a municipal project to equip all the households stuck in that situation with fiber.

You already have to download several gigabytes of stuff to run philomena so it's weird trying to save 20 mb

Meow:
Rebuilds. I do them a lot.

PR #580, opened 2025-05-15
Post-refactoring and improving of the new batch tags update flow

MareStare:
There are several changes packed in this single PR

## Zero changes edge case

Fixed this edge case when the user doesn't select any tags to revert at all and clicks on the "Revert selected" button:
image

(this bug should active be in prod too).

## Eliminate redundant preloads

I removed the redundant `:tag` and `:tag_change` (reverse) preloads for the initial `TagChange` select. The inside `Images.batch_update` doesn't need the full `TagChange.Tag` struct, it can simply work with tag_ids lists.

## More info in the flash

Also added the number of actually updated tags to the result of the revert:
image

## TODO

I'll try to add some auto tests for this code


Meow:
You should've asked me why there was mass_revert_tags... I intend to implement per-tag reverts as a mod tool in the future, so that mods can select individual tags to be reverted as opposed to whole batches.

Meow:
Split refactoring / logic fixes, and unit tests, into two separate PRs, this PR is becoming way too large.

PR #592, opened 2025-05-20
Expose Postgres port in docker-compose for connecting via a DB viewer client

MareStare:


Meow:
What's wrong with `psql`

MareStare:
> What's wrong with psql?

`psql` is TUI, while I prefer a GUI solution (especially built-in to VSCode) that always shows you the list of tables on the left that ou can quickly navigate through, provide syntax highlighting, autocompletion, error squiggles etc.

I've changed the default port number for postgres to 5400, with the override configurable via a `.env`

MareStare:
Just want to say, let's not overthink this. This PR was intended as a single line change to allow connecting to the db from an external viewer client. Using a. On 5432 port is fine as long as no developer has anything constantly running on 5400. I could expose the port under a compose profile though, but I don't want to add smth like pgamin to docker compose now.

Would that work for you?

MareStare:
The multi-compose file solution is described here: https://stackoverflow.com/a/73065022

PR #593, opened 2025-05-21
Fix reverts of self-canceling and non-current sequences, show effective tag changes on UI

MareStare:
This PR depends on this refactoring, so review+merge it first: https://github.com/philomena-dev/philomena/pull/594

Example that this fixes is below. I reverted the first and the second change, which resulted only in `tag1` being removed because `tag2` and `tag3` were already self-cancelled.

image

The new logic goes as this:
1. Accept a map `{ tag_change_id => [tag_ids] | nil }`. 
2. If the entry is `nil` - it means a request to revert changes to all tags in that `TagChange`, otherwise only revert the specified `tag_ids` within that `TagChange`.
3. For this input, fetch the `{image_id, tag_id, last_change: [change_id, added: boolean]}` from the DB. Filter out any change where the first change and the last change to the tag are not equal (i.e. self-cancelling changes). *The important nuance* here is that we don't need to fetch intermediate tag changes! We only need to know the first and the last change in the sequence to effectively revert it.
4. Fetch the IDs of the latest changes for each image/tag in the input.
5. Keep only changes where the `last_change` is the actual last change that determined the current state of the tag on the image.
6. Prepare the `added/removed` tag IDs for the image batch update logic

Also fixes this error when zero tag changes are selected:

### DIFF
--- a/lib/philomena/images.ex
+++ b/lib/philomena/images.ex
@@ -1074,13 +1074,11 @@ defmodule Philomena.Images do
 
     to_insert =
       Enum.flat_map(changes, fn change ->
-        Enum.map(change.added_tags, &%{tag_id: &1.id, image_id: change.image_id})
+        Enum.map(change.added_tag_ids, &%{tag_id: &1, image_id: change.image_id})
       end)
 
     to_delete_ids =
-      Enum.flat_map(changes, fn change ->
-        Enum.map(change.removed_tags, & &1.id)
-      end)
+      Enum.flat_map(changes, & &1.removed_tag_ids)
 
     to_delete =
       Tagging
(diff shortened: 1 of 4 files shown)


lib/philomena/images.ex
```diff
@@ -1166,21 +1180,21 @@ defmodule Philomena.Images do
...
-        |> Enum.flat_map(& &1.removed_tags)
-        |> Enum.uniq_by(& &1.id)
+        |> Enum.flat_map(& &1.removed_tag_ids)
+        |> Enum.uniq()
 
       %{
         image_id: image_id,
-        added_tags: Enum.reject(added, fn a -> Enum.any?(removed, &(&1.id == a.id)) end),
-        removed_tags: removed
+        added_tag_ids: Enum.reject(added, fn a -> Enum.member?(removed, a) end),
```
mdashlw, line 1193: `added -- removed`

MareStare: Didn't know of this operator, thanks

PR #594, opened 2025-05-21
Introduce a "Principal" type spec

MareStare:
It's really hard to express how confusing it is to see a function signature like `create_image(attribution, attrs)`. In fact, that `attribution` denotes the `principal` performing the action. In the long run all such confusing signatures should be refacotred, but for now I'm just introducing the type annotation for them to ease reading that code for a bit

### DIFF
--- a/lib/philomena/images.ex
+++ b/lib/philomena/images.ex
@@ -37,6 +37,7 @@ defmodule Philomena.Images do
   alias Philomena.Galleries.Gallery
   alias Philomena.Galleries.Interaction
   alias Philomena.Users.User
+  alias Philomena.Users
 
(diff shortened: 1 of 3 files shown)


lib/philomena/images.ex
```diff
@@ -69,6 +70,16 @@ defmodule Philomena.Images do
     |> Enum.map_join(", ", & &1.name)
   end
 
+  @typedoc """
+  Result of the `create_image/3` function. The image was created in a DB but
+  an upload process is probably still running in the background with its PID
```
Meow, line 75: "is probably still running" -> "could still be running"

MareStare: Fixed

PR #611, opened 2025-07-18
Strict approval

mdashlw:
- Flags all posts and comments with external links from untrusted users for review
- Trusted means a verified account or an account older than 2 weeks
- Approval logic for conversation messages remains the same (flag image embeds)
- Checks users About Me and Personal Title for external links and creates system reports

### DIFF
--- a/lib/philomena/schema/approval.ex
+++ b/lib/philomena/schema/approval.ex
@@ -1,41 +1,42 @@
 defmodule Philomena.Schema.Approval do
-  alias Philomena.Users.User
   import Ecto.Changeset
 
   @image_embed_regex ~r/!+\[/
 
-  def maybe_put_approval(changeset, nil),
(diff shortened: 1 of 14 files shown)


lib/philomena/schema/approval.ex
```diff
@@ -1,41 +1,45 @@
...
   import Ecto.Changeset
 
   @image_embed_regex ~r/!+\[/
 
-  def maybe_put_approval(changeset, nil),
-    do: change(changeset, approved: true)
+  defp external_link_regex do
+    site_domains =
+      (String.split(Application.get_env(:philomena, :site_domains), ",") ++
+         [Application.get_env(:philomena, :cdn_host), Application.get_env(:philomena, :camo_host)])
```
mdashlw, line 9: Now that I think about it, maybe Camo host shouldn't be on the whitelist

PR #617, opened 2025-07-30
Replace Jason

mdashlw:
Move to https://hexdocs.pm/elixir/JSON.html . Performance negligible


Meow:
For some context: the stdlib `JSON` module was designed and written by the Jason author under collaboration with Elixir maintainers. From synthetic benchmarks, it's slightly faster than Jason, and is supposed to have no difference in functionality / API.

PR #620, opened 2025-09-07
Autocomplete properties on query fields

koloml:
### Before you begin

---

Autocomplete now suggests properties!

image

## Changes

- Added properties autocomplete for:
  - Images
  - Tags
  - Forum
  - Comments
  - Reports
  - Filters
- Properties are suggested by default but they can be disabled with new additional setting
  image
- Added `window.booru.userRole` property containing current role of the user if there is one.
  - Some properties will only be suggested to staff members and if `hideStaffTools` is not enabled in the settings
- Some properties will also suggest available values
  - For example, `my:` will suggest `my:watched`, `my:upvotes`, `my:faves`, etc.
    image
  - Properties `implies:`, `implied_by:`, `aliases:` and `alias_of:` in Tags search will invoke local autocomplete and will try to suggest tags
    image
  - Numeric and date properties like `score:` will also suggest relevant operators like `score.gt:`, `score.lt:`, etc.
    image


Meow:
is merged now

PR #637, opened 2025-10-29
Convert the remaining JS files to Typescript

Meow:
- Converts all the remaining files from JS to Typescript
- Adds vitest tests for these files with 100% coverage
- Moves type definitions from the `types/` folder to their respective files, since the `types/` folder was merely a workaround for some files still being JS
- Consolidates `app.ts` and `when-ready.ts` together for organization
- Changes the behavior of `whenReady` function to accept spread syntax (array) of functions, so that it's testable via DOM tests
- Adjusts vitest excludes to not test coverage on files in `__tests__` folder, as well as `app.ts`
- Adjusts vitest config to require 100% coverage on the `js/query` folder
- Eliminates the usage of `el.querySelector` and `el.querySelectorAll` in favor of `$` and `$$`
- Renames files to use `-` instead of `_` in the filename for consistency
- Moves `tagsVersion` and `hiddenTag` to constants exported by `booru.ts` instead of a variable of `window.booru`


assets/js/app.ts
```diff
@@ -1,13 +1,85 @@
...
+      fn();
+    } catch (err: unknown) {
+      console.log(`${fn.name} ran with errors.`);
+
+      if (err instanceof Error) {
+        console.log(`The error was:\n\n${err.message}`);
+      }
+    }
+  });
+});
```
liamwhite: This logic should be moved into whenReady.

e.g., whenReady should be declared as
`export function whenReady(...callbacks: VoidFunction[]) { for (const callback of callbacks) { ...  } }`

Then app.ts can simply do:

```ts
whenReady(
  loadBooruData,
  listenAutocomplete,
  // ...
);
```

That way this module (app.ts) will end up having only one testable source line, so excluding it from coverage becomes more reasonable.

assets/js/boorujs.ts
```diff
@@ -0,0 +1,214 @@
...
+
+// Use this function to apply a callback to elements matching the selectors
+function selectorCb(base: ParentNode = document, selector: string, cb: (el: Element) => void) {
+  $$(selector, base).forEach(cb);
+}
+
+function selectorCbChildren(base: ParentNode = document, selector: string, cb: (el: Element) => void) {
+  const sel = $$(selector, base);
+
+  for (const el of Array.from(base.children)) {
```
liamwhite: Array.from shouldn't be needed I think, it should just let you iterate the children list

assets/js/image_expansion.ts
```diff
@@ -169,10 +210,11 @@ function bindImageTarget(node = document) {
...
     bindImageForClick(target);
+
     window.addEventListener('resize', () => {
       pickAndResize(target);
     });
   });
 }
 
-export { bindImageTarget };
+export { bindImageTarget, selectVersion, pickAndResize };
```
liamwhite: Change to export by function

assets/js/booru.ts
```diff
@@ -263,7 +263,4 @@ export function loadBooruData() {
   }
 }
 
-window.booru = {
-  hiddenTag: hiddenTagPath,
-  tagsVersion,
-} as BooruObject;
+window.booru = {} as BooruObject;
```
liamwhite: imagesclientside.ts needs to be adjusted to import the new constant hiddenTagPath

liamwhite (approved):

PR #638, opened 2025-10-30
Bump minimum browser versions to Sep. 2019 browsers

Meow:
September 2019 is roughly 6 years ago, that's plenty of time ago.

I plan to bump these with every time Apple makes a new iOS tbh.

### DIFF
--- a/assets/.browserslistrc
+++ b/assets/.browserslistrc
@@ -1,8 +1,10 @@
-last 2 Android versions
-last 2 Chrome versions
-last 2 ChromeAndroid versions
-last 2 Edge versions
-last 2 Firefox versions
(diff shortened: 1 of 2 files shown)


Meow:
> Did this enable anything we want to use

Most of the things were already available in the previous browser list. The one interesting thing that was enabled was `Object.fromEntries` which required Chrome 73 or newer, which is required by certain frameworks such as React.

Other than that, JS operators `?.` and `??` are now supported.

When we bump the minimum to iOS 14 next year (as per my proposed plan), it'll enable many interesting things related to CSS, such as `gap` in flexboxes, which we already started using (maybe wrongfully so, actually).

PR #673, opened 2026-02-11
Better ICC profile handling in image processing

### LINKED ISSUE #665: EXIF and ICC profiles not stripped from PNGs
Just not a thing. It should be to be consistent with JPEG processing. Needs auto-orientation, stripping, and profile conversion.

https://github.com/philomena-dev/philomena/blob/b7a95ee070036b5be3bd5f5ba5ee26b58c04aadb/lib/philomena_media/processors/png.ex

### LINKED ISSUE #666: Smarter ICC profile handling
Some JPEG images embed an sRGB ICC profile, which we remove by converting it to sRGB, resulting in unnecessary quality loss.

https://github.com/philomena-dev/philomena/blob/b7a95ee070036b5be3bd5f5ba5ee26b58c04aadb/lib/philomena_media/processors/jpeg.ex

There's no nice way to determine if two profiles are effectively the same.

Idea: extract the profile and process a sample image with it to see if it matches our reference sRGB or If the target image needs to be processed further. 

Simpler idea: skip re-encoding for all profiles with "sRGB" in the name.

mdashlw:
Closes #666 and closes #665

- Updates imagemagick inside mediaproc to v7 Q16 [HDRI](https://imagemagick.org/script/high-dynamic-range.php) - necessary for accurate color space conversion
- PNG: always runs non-animated files through magick reorientation, profile conversion, and metadata stripping
- JPEG: checks the embedded profile against our reference first to avoid reencoding 

Optimal reference image with RGB data made:
```
magick -size 1x1 xc:"rgb(16,16,16)" xc:"rgb(128,128,128)" xc:"rgb(128,0,0)" xc:"rgb(0,128,0)" xc:"rgb(0,0,128)" xc:"rgb(64,128,192)" +append -depth 8 -type TrueColor -strip priv/icc/reference.png

magick priv/icc/reference.png -depth 8 RGB:- > priv/icc/reference.rgb
```


liamwhite (approved):

Meow:
Fix the sobelow warning and I'll merge this

```
##############################################
#                                            #
#          Running Sobelow - v0.14.1         #
#  Created by Griffin Byatt - @griffinbyatt  #
#     NCC Group - https://nccgroup.trust/     #
#                                            #
##############################################

Traversal.FileModule: Directory Traversal in `File.read` - Low Confidence
File: lib/philomena_media/icc.ex
Line: 30
Function: srgb_profile?:15
Variable: reference_rgb


def srgb_profile?(file) do
  profile = Briefly.create!(extname: ".icc")

  with {_output, 0} <- Remote.cmd("magick", [file, profile]),
       {test, 0} <-
         Remote.cmd("magick", [
           reference_image(),
           "-profile",
           profile,
           "-profile",
           srgb_profile(),
           "-depth",
           "8",
           "RGB:-"
         ]),
       {:ok, reference} <- File.read(reference_rgb()) do
    test == reference
  else
    _ -> false
  end
end

-----------------------------------------------

... SCAN COMPLETE ...
```

liamwhite:
@Meow fixed

lib/philomena_media/strip.ex
```diff
@@ -0,0 +1,36 @@
...
+  alias PhilomenaMedia.Icc
+  alias PhilomenaMedia.Remote
+
+  @spec requires_strip?(Path.t()) :: boolean()
+  def requires_strip?(file) do
+    with {output, 0} <-
+           Remote.cmd("magick", ["identify", "-format", "%[orientation]\t%[profile:icc]", file]),
+         [orientation, profile] <- String.split(output, "\t") do
+      orientation not in ["Undefined", "TopLeft"] or
+        (profile != "" and not Icc.srgb_profile?(file))
```
uis246, line 13: Wait. It strips non-sRGB profiles, but keeps if they are?

liamwhite: Yes - this avoids re-encoding JPEG files with colorspaces that are equivalent to sRGB, but otherwise re-encodes to normalize to sRGB

uis246:
I hope it's my brain being stupid and seeing things, and it doesn't strip non-sRGB profiles.

lib/philomena_media/strip.ex
mdashlw: Ideally it'd still strip sRGB-like profiles without re-encoding. Did I forget to implement that or is that handled somehow?

uis246: Forgot I did not continue my comment here. What I wanted to ask is if images in original space can still be downloaded. And what tonemapping algorithm is used in conversion? Just clamping to unorm in sRGB can produce hotspots of solid colour in places, where there is gradient that overflows normalised sRGB.

PR #684, opened 2026-05-01
Expose image subscriptions similar to hides (#527).

luckydonald:
As the count of hides of an image are provided,
similar shall be the (comment-)subscription count.

This is based on the changes for `hides` in ba904786de8f34ce8729759a68622779584840bd (know as eef9ec612fc19bd39bdb09e0a354f5a8d4590db8 after rebase).

---

My 

---

Changes made:

1. Migration: adds `subscriptions_count integer not null default 0` to images and backfills it from `image_subscriptions`.
2. `image.ex`: added `field :subscriptions_count, :integer, default: 0`.
3. `subscriptions.ex`: added `defoverridable create_subscription: 2, delete_subscription: 2` at the end of the `quote` block so callers can override with `super/2`.
4. `philomena/images.ex`: added `create_subscription/2` and `delete_subscription/2` overrides that call `super` then increment/decrement `subscriptions_count` atomically via `Repo.update_all(inc: ...)`.
5. `search_index.ex`: added `subscriptions: %{type: "integer"}` to the mapping and `subscriptions: image.subscriptions_count` to `as_json/1`.
6. `query.ex`: added `subscriptions` to `int_fields` in `anonymous_fields/0`.
7. `image_sorter.ex`: added `subscriptions` to `@allowed_fields`.
8. `_form.html.slime`: added `"Sort by subscriptions": :subscriptions` after the `hides` entry.
9. `json/image_view.ex`: added `subscriptions: image.subscriptions_count` to the image JSON response.

### Before you begin

---

### DIFF
--- a/lib/philomena/images.ex
+++ b/lib/philomena/images.ex
@@ -43,6 +43,26 @@ defmodule Philomena.Images do
     on_delete: :clear_image_notification,
     id_name: :image_id
 
+  def create_subscription(object, user) do
+    result = super(object, user)
+
+    if match?({:ok, _}, result) do
+      Image |> where(id: ^object.id) |> Repo.update_all(inc: [subscriptions_count: 1])
+    end
+
+    result
+  end
+
+  def delete_subscription(object, user) do
+    result = super(object, user)
+
+    if match?({:ok, _}, result) do
+      Image |> where(id: ^object.id) |> Repo.update_all(inc: [subscriptions_count: -1])
+    end
+
+    result
+  end
+
   @doc """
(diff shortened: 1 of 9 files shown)


liamwhite:
What benefit does exposing the subscription count have?

luckydonald:
@liamwhite Mostly symmetry to the `hides`, while I worked on #685.

Meow:
I'll discuss this change with the booru mod team. They're ultimately the ones who decide what goes policy-wise, and exposing an extra number seems like a policy decision.

PR #685, opened 2026-05-01
Allow to search for `my:subs`

luckydonald:
This is to allows to have some way to list the images you have subscribed to,
which does not exist currently.

While add it, also added the missing `my:hidden` to autocomplete.

### Before you begin

---

### DIFF
--- a/assets/js/autocomplete/__tests__/properties/values.spec.ts
+++ b/assets/js/autocomplete/__tests__/properties/values.spec.ts
@@ -23,6 +23,7 @@ autocompleteTest('should suggest values for some properties', async ({ ctx, expe
       "suggestions": [
         "(property) my:comments",
         "(property) my:faves",
+        "(property) my:hidden",
         "(property) my:subs",
         "(property) my:uploads",
         "(property) my:upvotes",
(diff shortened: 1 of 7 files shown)


luckydonald:
using `my:subscriptions` would be possible, too, but with stuff like `favorites` shortened to `faves`, I think shorter is better.

Meow:
~~If I have my subs, does that mean there are also my doms?~~

lgtm, seems like you also fixed some oddities with my:hides and my:queries as anon, if @liamwhite has no concerns about performance I'd say this is good to merge

liamwhite:
I have no objections regarding performance

luckydonald:
From the timeline of the merge this should have been included in `1.2.4`, right?

image

Asking first, before I go down a witch hunt if it is not yet deployed.

liamwhite:
No, it wasn't in 1_2_stable

luckydonald:
@liamwhite This explains a lot, thanks.
Any ETA by chances, or insights how the release of that would happen?

liamwhite:
I'm not in charge of deployments or releases so I can't really answer that

PR #687, opened 2026-05-02
Add autocomplete history item deletion button

MareStare:
This PR adds a little `xmark` to history items, right-aligned with image counts. It is highlighted with a color change on hover, and clicking it triggers a deletion of the history item. The change is immediately persisted to local storage and the suggestions are refreshed without closing the popup.

(this recording is outdated, see the latest version in PR comments)

https://github.com/user-attachments/assets/78e81380-b3e7-42c4-8969-5a3e18cf6f7c

### DIFF
--- a/assets/css/themes/base/dark.css
+++ b/assets/css/themes/base/dark.css
@@ -177,6 +177,7 @@
 
     --autocomplete-history-color: $autocomplete-history-color;
     --autocomplete-history-match-color: hsl(from $autocomplete-history-color h s calc(l + 20));
+    --autocomplete-history-delete-hover-background: hsl(from $autocomplete-history-color h s l / 0.25);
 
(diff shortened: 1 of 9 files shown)


liamwhite:
Would it make sense for the deletion icon to be visually distinguished in a different way like color, so as to not appear as being a link?

MareStare:
> Would it make sense for the deletion icon to be visually distinguished in a different way like color, so as to not appear as being a link?

I did try to give it a different color, but, IDK, it just starts to stand out too much, when there shouldn't be any focus on this button at all. It will be used rather rarely. But I'm open to any better ideas for the style of the button

image

Chrome uses the same white colour for the button as for the text of the item
image

MareStare:
I think this variant looks a little better. The background changes on hover, which makes it look a bit more like a button. Also made the `X` a little smaller to deemphasize it.

image

liamwhite (approved):

PR #693, opened 2026-05-05
Strip degenerate animated PNGs

liamwhite:
The current processing code inconsistently handles animated PNG with 1 frame. This adds a normalization pass to strip these files to still PNG.


uis246:
~~Nice getting rid of degenerates~~
Getting rid of acTL and fcTL chunks should be enough without invoking reencoding.

PR #703, opened 2026-06-07
Minor CI improvements

liamwhite:
Primarily, makes the actions cache actually work (eliminates repeated dep compilation and PLT rebuilding on Dialyzer runs), but also various CI maintenance items.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,7 @@ jobs:
       - run: docker compose pull
       - run: docker compose build
       - run: ./scripts/philomena.sh test
+      - run: sudo chown -R $USER $GITHUB_WORKSPACE
 
   typos:
     name: 'Check for spelling errors'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,8 +18,12 @@ jobs:
 
       - run: docker compose pull
       - run: docker compose build
-      - run: ./scripts/philomena.sh test
-      - run: sudo chown -R $USER $GITHUB_WORKSPACE
+      - name: Run tests
+        run: ./scripts/philomena.sh test
+      # Artifacts from docker bind mount are owned by root, but cache action
+      # can only save artifacts running under the default user:
+      - name: Change artifact ownership for cache action
+        run: sudo chown -R $USER $GITHUB_WORKSPACE
 
   typos:
     name: 'Check for spelling errors'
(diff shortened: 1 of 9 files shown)


docker/app/Dockerfile
```diff
@@ -31,13 +31,11 @@ RUN apk add \
 # In any case, it's much better to use rustup for rust toolchain management.
 # Also, this lives in a separate RUN statement to cache this step separately, as
 # it's quite slow to install.
-RUN --mount=source=./scripts/lib.sh,target=./scripts/lib.sh \
-    --mount=source=./scripts/install/rust.sh,target=./scripts/install/rust.sh \
-    --mount=source=./rust-toolchain.toml,target=./rust-toolchain.toml \
-    ./scripts/install/rust.sh
+COPY scripts scripts
```
MareStare, line 32: The comment above gives some reasoning to keep rust installation in a separate RUN statement. So let's either follow that comment or remove it as it became stale.

docker/app/Dockerfile
```diff
@@ -31,13 +31,11 @@ RUN apk add \
...
+COPY rust-toolchain.toml rust-toolchain.toml
 
-RUN --mount=source=./scripts,target=./scripts \
```
MareStare, line 39: Not sure what the problem is with a build mount. This way scripts used only during build don't end up in the final image, which anyway uses a volume mount of the repo workspace at the same location and thus shadows these scripts

.github/workflows/ci.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       - run: docker compose pull
       - run: docker compose build
       - run: ./scripts/philomena.sh test
+      - run: sudo chown -R $USER $GITHUB_WORKSPACE
```
MareStare: What's the reason to do a chown at the end? This code runs on an emphemeral VM, so it seems redundant.

UPD, I guess it's for the cache action, in which case a comment about that would be nice

docker/app/Dockerfile
liamwhite: It comes from a [previous attempt to use buildx with GHA cache](https://github.com/philomena-dev/philomena/commit/ad3766f2e8fb2dd658b9ec511fe2084f06cd3542) in which having a build-time bind mount present completely prevents caching the layer (since the inputs to the build are no longer static and can't be digested)

MareStare: Makes sense

.github/workflows/ci.yml
liamwhite: Comment added

docker/app/Dockerfile
liamwhite: Edited the comment

MareStare (approved):

PR #707, opened 2026-07-07
OpenAPI Spec for API v1

Meow:
Generated by Claude Fable 5 after looking at the actual controllers + derpi's API docs page

I wrote maybe 5% of this file (the topmost description + site links + small corrections). Think of it what you will.

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -1,0 +1,2014 @@
+openapi: 3.1.1
+info:
+  title: Philomena API
+  description: |
+    JSON API for major site functionality of the Philomena imageboard software
+    (as used by Derpibooru and others). The API can be freely used by anyone
+    wanting to produce tools for the site or other web applications that use
+    the data provided within the site.
+
+    Note that if you are looking to *continuously scrape the entire website*,
+    several boorus offer nightly database dumps instead. Consider if these suit
+    your needs first, then rely on the API if they do not.
+
+    **Data Dumps:**
+
+    - [Derpibooru](https://derpibooru.org/pages/data_dumps)
+    - [Furbooru](https://furbooru.org/pages/data_dumps)
+    - [Tantabus](https://tantabus.ai/pages/data_dumps)
+
+    ## Licensing
+
+    Anyone may use the API. Users making abusively high numbers of requests or
+    excessively expensive requests will be asked to stop, and banned if they
+    do not. Your application must properly cache, and respect server-side
+    cache expiry times. Your client must gracefully back off if requests fail,
+    preferably exponentially or fatally.
+
+    If images are used, the artist must always be credited (if provided) and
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -0,0 +1,2014 @@
...
+      tags: [Images]
+      summary: Upload an image
+      description: |
+        Submits a new image. Both `key` and `url` are required when posting
+        via the JSON API; alternatively (on self-hosted instances) the file
+        may be sent as a direct multipart upload in `image[image]`.
+
+        You *must* provide the direct link to the image in the `url`
+        parameter, and you *must* set the `content-type` header to
+        `application/json` for the site to process your request.
```
mdashlw: This is not technically true. The API supports `multipart/form-data` and uploading the image file directly under `image[image]`.

openapi.yaml
```diff
@@ -0,0 +1,2014 @@
...
+            description:
+              type: string
+              maxLength: 50000
+              description: Image description, in Markdown.
+            tag_input:
+              type: string
+              description: >
+                Comma-separated list of tags. At least three tags are
+                required, including a rating tag.
+            source_url:
```
mdashlw, line 1286: Multiple sources technically supported using this weird syntax:
```json
{"image":{"sources":{"0":{"source":"source 1"}}}}
```

openapi.yaml
Meow: @mdashlw how about

> Submits a new image. Both `key` and `url` are required when posting
> via the JSON API; alternatively (on self-hosted instances) the file
> may be sent as a direct multipart upload in `image[image]`.
> 
> **URL-based uploads:**
> 
> Set `content-type` to `application/json`. The following fields are required in the JSON body:
> 
> ```json
> {
>   "image": {
>     "tag_input": "safe, cute, fox", // at least 3 tags (5 on Furbooru), must include a rating tag
>   },
>   "url": "https://example.com/cute-fox.png" // direct links only
> }
> ```
> 
> **Multipart Form Uploads:**
> 
> Set `content-type` to `multipart/form-data` and send a multipart body with the following fields (Javascript example):
> 
> ```javascript
>   const formData = new FormData()
>   formData.append('image[image]', '@filename')
>   formData.append('image[tag_input]', 'safe, cute, fox') // at least 3 tags (5 on Furbooru), must include a rating tag
> 
>   fetch('/api/v1/json/images?key=YOUR_SECRET_TOKEN', {
>     method: 'POST',
>     headers: {
>       'Content-Type': 'multipart/form-data'
>     },
>     body: formData
>   })
> ```
> 
> At least three tags must be provided via `image[tag_input]`, including
> a rating tag.
> 
> You are *strongly recommended* to test code using this endpoint using
> a local copy of the website's source code. Abuse of the endpoint
> **will result in a ban**.

PR #723, opened 2026-07-13
GitHub/VSCode Style Diffs

Meow:
Changes diffs to be per-line, highlights differences between the lines, displays only 3 previous/next lines for context, hides the rest of the unchanged lines.


test/philomena_web/controllers/rule_controller_test.exs
```diff
@@ -86,6 +86,31 @@ defmodule PhilomenaWeb.RuleControllerTest do
...
+
+      conn = get(conn, ~p"/rules/#{rule}")
+      response = html_response(conn, 200)
+
+      # The description renders as a line-by-line unified diff table over the
+      # raw markdown source. The cell text is the escaped source, so the
+      # unchanged suffix "rule text" appears literally in the diff__text cell
+      # rather than inside rendered markup. (The two versions share a
+      # same-second timestamp, so the diff direction is not fixed - assert on
+      # markup that holds either way.)
```
liamwhite, line 105: !?, not related to this change but definitely a bug that should be fixed - versions should be serialized and diffs should be deterministic

native/philomena/src/markdown_diff.rs
```diff
@@ -0,0 +1,186 @@
+//! GitHub-style "pretty" diffing of two Markdown documents.
+//!
```
liamwhite: does this //! comment mean I am supposed to pay special attention? I've never seen it used before to indicate anything other than "this is probably a bug"

native/philomena/src/markdown_diff.rs
```diff
@@ -0,0 +1,186 @@
...
+//! rows are marked with `diff__row--del` / `diff__row--ins`, and runs of
+//! unchanged lines beyond the context window collapse into an
+//! "N unchanged lines" separator row. When a line is edited in place, the
+//! exact changed words inside it are additionally wrapped in
+//! `<del class="diff__hl">` / `<ins class="diff__hl">`.
+//!
+//! The source text is HTML-escaped as it is emitted, so untrusted input stays
+//! inert; the only live markup is the table structure and the diff wrappers.
+
+use similar::{Algorithm, ChangeTag, InlineChange, TextDiff};
```
liamwhite, line 14: from similar crate-doc:
> For large and very distinct inputs the algorithms as implemented can take a very, very long time to execute. Too long to make sense in practice. To work around this issue all diffing algorithms also provide a version that accepts a deadline which is the point in time as defined by an Instant after which the algorithm should give up. 

this should almost certainly be implemented to prevent diff DoS

native/philomena/src/markdown_diff.rs
MareStare: It's just Rust's module-level ["inner" doc comment syntax](https://doc.rust-lang.org/reference/comments.html#examples)

liamwhite (approved):

PR #724, opened 2026-07-16
Fix static page diffs & make diff direction deterministic

Meow:
Fixes the static page diffs off-by-one issue, as well as non-determinism in static page and rule diffs when the versions are timestamped the exact same datetime, now sorts by both creation date and uses the version's ID as a source for determinism.


lib/philomena/rules.ex
liamwhite, line 1: In rule_controller_test.exs, there's a comment about rule version ordering being non-deterministic which should be removed now that it is.

liamwhite (approved):

PR #726, opened 2026-07-16
async mime type processing in mediaproc

Meow:
replaces invokations of `file -b --mime-type` in elixir code


liamwhite (approved):
I will refactor to extract shared pipeline logic before doing the rest of these calls

PR #727, opened 2026-07-16
Introduce a search index versioning and migration mechanism

Meow:
Allows search indices to be updated without deleting and recreating the whole schema every time. Zero-downtime migrations too.

### DIFF
--- a/lib/philomena_query/search.ex
+++ b/lib/philomena_query/search.ex
@@ -11,6 +11,7 @@ defmodule PhilomenaQuery.Search do
 
   alias PhilomenaQuery.Batch
   alias PhilomenaQuery.Search.Api
+  alias PhilomenaQuery.Search.WriteTargets
   alias Philomena.Repo
   require Logger
   import Ecto.Query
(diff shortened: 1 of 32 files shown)


lib/philomena_query/search/write_targets.ex
```diff
@@ -0,0 +1,168 @@
...
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  @doc """
+  Return the index names a document write for `alias_name` must target.
+
+  With a single physical index holding the alias (or a bare concrete index
+  occupying the alias name), this is `[alias_name]`, producing requests
+  identical to today's. With more than one member - a migration in progress -
```
liamwhite: "today's"

lib/philomena_query/search.ex
```diff
@@ -92,46 +93,191 @@ defmodule PhilomenaQuery.Search do
...
+
+  - `alias:` - the (prefixed) alias name
+  - `status:` - `:aliased` when a physical index holds the alias, `:legacy`
+    when a bare concrete index occupies the alias name, `:missing` otherwise
+  - `live_physical:` - the index currently serving reads, if any
+  - `live_version:` - from the live index's `mappings._meta.version`, falling
+    back to the `_v` name suffix; `nil` for legacy indices without either
+  - `live_mapping:`, `live_settings:` - the live index's definition, for
+    `PhilomenaQuery.Search.MappingDiff`
+  - `orphans:` - members which are not the live index (stale migration
```
liamwhite:
```suggestion
  - `alias` - the (prefixed) alias name
  - `status` - `:aliased` when a physical index holds the alias, `:legacy`
    when a bare concrete index occupies the alias name, `:missing` otherwise
  - `live_physical` - the index currently serving reads, if any
  - `live_version` - from the live index's `mappings._meta.version`, falling
    back to the `_v` name suffix; `nil` for legacy indices without either
  - `live_mapping`, `live_settings` - the live index's definition, for
    `m:PhilomenaQuery.Search.MappingDiff`
  - `orphans` - members which are not the live index (stale migration
```

lib/philomena_query/search.ex
```diff
@@ -216,17 +376,24 @@ defmodule PhilomenaQuery.Search do
   Note that indexing is near real-time and requires an index refresh before the document will
   become visible. Unless changed in the mapping, this happens after 5 seconds have elapsed.
 
+  During an index migration the write fans out to every physical index backing
+  the alias; a `targets:` option overrides resolution. Failures against
+  individual physical indices are tolerated (a concurrently deleted old index
+  responds 404 while the surviving copy succeeds).
```
liamwhite, line 381:
```suggestion
  During an index migration the write fans out to every physical index backing
  the alias; resolution can be overridden with the `targets` option. Failures against
  individual physical indices are tolerated (a concurrently deleted old index
  responds 404 while the surviving copy succeeds).
```

liamwhite (approved):
Finally catching up to year 2015 technology

PR #729, opened 2026-07-17
Move user settings columns to their own table

Meow:
Move any columns responsible for user settings (stuff that affects how the user sees the site, not how other users see the user) to their own table.

TODO: drop old columns


liamwhite (approved):
Very natural seam to pull self-selected preferences from the rest of the user schema

PR #734, opened 2026-07-18
Fix self-cancelling batch reverts

### LINKED ISSUE #585: Fix the batching logic of the full tag changes revert worker
The logic of splitting the full-revert into batches has a logic bug. Suppose there was a sequence:
1. `+tag`
2. `-tag`

The full revert of this sequence should resolve to a noop since this sequence already canceled itself. But... If we do reverts in a batches of N items, and the batch is split in the middle of this sequence, then:

The first batch will try to revert `+tag`, it'll see that this isn't the current change and ignore it.
The second batch will try to revert `-tag` and it'll do that - it'll add the `tag` to the image.

So we just can't split tag changes into batches for reverts. I'm sure there is a mathematical term for this phenomena, but the gist is that logic breaks if you split it into batches. The best we can do is batch changes only across the (image_id, tag_id) dimension.

It also doesn't mean the +tag and -tag need to be strictly adjacent. There can be hundends of other tag changes between them:
+tag,
N. ... (hundereds of other tag changes that don't modify tag)
N + 1. -tag

I think this best explains why this logic is unbatchable across TagChange dimension, especially because as it's implemented today tag changes from unrelated images can be batched together thereby increasing the probability of a broken batch split. 

## Implementation Notes

This logic lives here: 
https://github.com/philomena-dev/philomena/blob/b3efebcd8a443c3363aae6bacccbcf9223ad5024/lib/philomena/workers/tag_change_revert_worker.ex#L27-L35

Meow:
Fixes #585

### DIFF
--- a/lib/philomena/workers/tag_change_revert_worker.ex
+++ b/lib/philomena/workers/tag_change_revert_worker.ex
@@ -1,4 +1,9 @@
 defmodule Philomena.TagChangeRevertWorker do
+  @moduledoc """
+  Reverts every tag change made by a user, IP, or fingerprint, batching by
+  image so each image's tag history is reverted in a single operation.
+  """
+
   alias Philomena.TagChanges.TagChange
   alias Philomena.TagChanges
   alias PhilomenaQuery.Batch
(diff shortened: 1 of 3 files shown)


lib/philomena/workers/tag_change_revert_worker.ex
```diff
@@ -27,8 +32,10 @@ defmodule Philomena.TagChangeRevertWorker do
     batch_size = attributes["batch_size"] || 100
     attributes = Map.delete(attributes, "batch_size")
 
+    # Batch on image_id, never on tag change id: a batch boundary that splits
+    # one image's tag history would un-cancel a `+tag`/`-tag` pair
     queryable
-    |> Batch.query_batches(batch_size: batch_size)
+    |> Batch.query_batches(batch_size: batch_size, id_field: :image_id)
```
liamwhite, line 38: Unlike the primary key `tag_changes.id`, `tag_changes.image_id` is not a candidate key for the tag_change schema. I believe this means the results returned by `query_batches` will be ordered nondeterministically, which may have implications for how reverting works, and will break (not deadlock, but fail to return every instance of a result) if the number of instances for a given `image_id` value exceeds the batch size. Please check this

liamwhite (approved):

PR #735, opened 2026-07-18
Enforce tag length limits

### LINKED ISSUE #668: Tags need a name length limit
desperately

Meow:
Fixes #668 

Set to a fairly generous limit of 256 bytes. Subject to change.

Querying Derpibooru database reveals that `102` tags are over the 256 byte length. Looking at which ones exactly, I can confidently say, none of them are going to be missed.

EDIT: also fixed a bug where autocomplete binary was incorrectly capped to 254 byte instead of 255 byte tag names.

### DIFF
--- a/lib/philomena/autocomplete/generator.ex
+++ b/lib/philomena/autocomplete/generator.ex
@@ -139,7 +139,7 @@ defmodule Philomena.Autocomplete.Generator do
     # Sort is done in the application to avoid collation.
     tags =
       LocalAutocomplete.get_tags(@top_tags)
-      |> Enum.filter(&(byte_size(&1.name) < 255))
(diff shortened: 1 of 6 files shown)


mdashlw:
There was an idea before to make it 255 bytes specifically. Any reason against it?

Meow:
@mdashlw 256 seemed like a nice round number

liamwhite (approved):

mdashlw:
> @mdashlw 256 seemed like a nice round number

The justification was that 255 bytes is what the binary format of local autocomplete supports max:
https://github.com/philomena-dev/philomena/blob/82a138104b4a0b014fb9c4dcbfecc75074ce006e/lib/philomena/autocomplete/generator.ex#L138-L142

Note the byte vs character distinction too

PR #737, opened 2026-07-19
Get rid of rails polymorphic relations

Meow:
Because polymorphism is an anti-pattern


lib/philomena/reports.ex
```diff
@@ -82,15 +88,20 @@ defmodule Philomena.Reports do
...
+    type_and_id
+    |> new_report()
     |> Report.user_creation_changeset(attrs, attribution, rule)
     |> Repo.insert()
     |> reindex_after_update()
   end
 
+  defp new_report({reportable_type, reportable_id}) do
+    struct(Report, %{Report.column_for_type(reportable_type) => reportable_id})
+  end
```
liamwhite: A lot of this change design is based around having callsites continue to use the _type and _id columns, by adding them as virtual columns and adding a bunch of translation. I generally disagree and think that the callsites should be changed to reflect the required usage, because except for the listing/index (which you can already handle with pattern matching as the view code below does, or at worst, `cond` on which column is `not is_nil`), the callsites *already know* which column they are going to want.

Like here, you could have...
```
@spec create_report(attribution :: Keyword.t(), attrs :: map())
```
and then pass e.g. `comment_id` in attrs in the caller. Instead of needing to map type to the column to fill out.

lib/philomena/search_indexer.ex
```diff
@@ -158,15 +157,14 @@ defmodule Philomena.SearchIndexer do
   defp reindex_schema_impl(schema, opts)
 
   defp reindex_schema_impl(Report, opts) do
```
liamwhite: The special case for reports isn't needed anymore so this entire function body can be deleted and the one below merged into the head.

liamwhite (dismissed):

lib/philomena/reports.ex
```diff
@@ -71,34 +77,36 @@ defmodule Philomena.Reports do
...
+      iex> create_report([image_id: image.id], attribution, %{"reason" => "..."})
       {:ok, %Report{}}
 
-      iex> create_report(%{field: bad_value})
+      iex> create_report([image_id: image.id], attribution, %{"reason" => ""})
       {:error, %Ecto.Changeset{}}
 
   """
-  def create_report({reportable_type, reportable_id} = _type_and_id, attribution, attrs \\ %{}) do
+  def create_report(target, attribution, attrs \\ %{}) do
```
liamwhite: Flip the argument order so that you can use the condensed keyword syntax in the caller (comment_id: x) instead of ([comment_id: x])

(on all relevant functions)

lib/philomena/mod_notes/mod_note.ex
```diff
@@ -3,26 +3,110 @@ defmodule Philomena.ModNotes.ModNote do
...
-    field :notable_type, :string
+    belongs_to :user, User
+    belongs_to :report, Report
+    belongs_to :dnp_entry, DnpEntry
 
     field :body, :string
 
     field :notable, :any, virtual: true
+    field :notable_type, :string, virtual: true
+    field :notable_id, :integer, virtual: true
```
liamwhite: Guess I wasn't strong enough in my previous comment.

Eradicate these notable* virtual columns. You wouldn't write it this way if you were building the feature from a table that started off normalized. The code should not be stuck in this half-migrated state where it's still referencing the old design.

lib/philomena/reports/report.ex
```diff
@@ -18,15 +45,58 @@ defmodule Philomena.Reports.Report do
     field :open, :boolean, default: true
     field :system, :boolean, default: false
 
-    # fixme: rails polymorphic relation
-    field :reportable_id, :integer
-    field :reportable_type, :string
-
     field :reportable, :any, virtual: true
```
liamwhite: Similarly, stop using this `reportable` column.

liamwhite (approved):

PR #759, opened 2026-09-14
Make conversation list always sort unread-first (1.2 series)

liamwhite:
I can't really think of reason you would ever want this to not be the case so I'm not adding a toggle for it.


Meow (approved):
very good change, lgtm

PR #760, opened 2026-09-15
Erasure workflow optimization (1.2 series)

liamwhite:
Removes the superfluous confirmation prompts. The first prompt was never necessary, and the new erase page serves as its own confirmation.
Erase for spam is now optimized to four clicks: open profile page -> toggle staff tools -> target erase button at top of list -> target red "Erase user" button -> done
Three clicks if you use this userstyle which forces the staff tools section open:
```css
.js-admin__options__toggle.hidden {
  display: flex !important;
}
```


Meow (approved):
lgtm

PR #762, opened 2026-09-15
Port account flow fixes from updated phx.gen.auth (1.2 series)

liamwhite:
- Adds 7-day session/remember token rotation
  - Causes sessions not to expire with continuous use
- Adds two-step confirmation flow for account confirmation


Meow (approved):
lgtm

PR #763, opened 2026-09-16
Increase duplicate image diff contrast boost

avonder:
### Before you begin

---

The current value is proving to be insufficient and produce false negatives.


liamwhite:
Closing in favor of #764

